PR #23, opened 2015-04-28
Support for streamline

winniehell:
When using [streamline], source files end with `._js` instead of `.js`. This patch considers both extensions when searching for `require()`s.

[streamline]: /Sage/streamlinejs

### DIFF
--- a/index.js
+++ b/index.js
@@ -45,7 +45,7 @@ function checkDirectory(dir, ignoreDirs, deps, devDeps) {
   });
 
   finder.on("file", function (filename) {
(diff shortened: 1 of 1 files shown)


lijunle:
@winniehell Do you still need such feature now?

You can run `depcheck --parsers="*._js=es6"` after update to version 0.5.9.

Please let me know if you get any problems.

However, this RP is going to be closed because I don't think it could be merged.

winniehell:
@lijunle thank you for the answer! I'll try that. :+1:

PR #24, opened 2015-04-28
Speed up build

winniehell:
These settings may speed up the build as described [here](http://docs.travis-ci.com/user/workers/container-based-infrastructure/) and [here](http://docs.travis-ci.com/user/caching/).


lijunle:
Those have been a part of code now.

PR #25, opened 2015-04-28
Test coverage

winniehell:
* determine test coverage
* add missing test cases
* add error handler for walkdir

PR #27, opened 2015-06-18
Fix the typo: withouttDev -> withoutDev.

lijunle:


winniehell:
I found that one, too :wink: (7e7382fa988e1c369847d5de0b5bd39f8b6b98f5)
So it is already part of #25...

lijunle:
@winniehell Yes, after I sent out this PR. I found your fixes.

Besides, I think that the current assertion is not enough. When there is one unused package in the list, it should validate that element.

PR #30, opened 2015-07-31
Added possibility to whitelist modules

siimon:
I missed the functionality to whitelist a modules so I implemented it, let me know if you want me to change anything.


gtanner:
This is the same as ignored. Probably should opt to use that instead

lijunle:
Use `--ignores` parameter, which has been include d in version 0.5.9.

PR #35, opened 2015-10-12
Support for JSX files

jiripech:
Added one more extension of files we browse through and bumped a version number.


lijunle:
Version 0.5.9 has supported JSX natively. Please update to enjoy it.

PR #91, opened 2016-01-19
added an option to report on used dependencies

gtanner:
Added an optional param to report on the used dependencies as well.

Regular output will show a list of the used dependencies, the JSON output will show the files that those dependencies are also used in.

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -45,6 +45,7 @@ export default function cli(args, log, error, exit) {
     .describe('dev', 'Check on devDependecies')
     .describe('ignore-bin-package', 'Ignore package with bin entry')
     .describe('json', 'Output results to JSON')
+    .describe('used', 'Report on which dependencies are used')
     .describe('ignores', 'Comma separated package list to ignore')
     .describe('ignore-dirs', 'Comma separated folder names to ignore')
     .describe('parsers', 'Comma separated glob:pasers pair list')
@@ -83,7 +84,7 @@ export default function cli(args, log, error, exit) {
       detectors: getDetectors(opt.argv.detectors),
       specials: getSpecials(opt.argv.specials),
     }))
-    .then(result => output(result, log, opt.argv.json))
+    .then(result => output(result, log, opt.argv))
     .then(({ dependencies, devDependencies }) =>
       exit(opt.argv.json
         || dependencies.length === 0 && devDependencies.length === 0
--- a/src/index.js
+++ b/src/index.js
@@ -148,7 +148,7 @@ function checkFile(dir, filename, deps, parsers, detectors) {
     getDependencies(dir, filename, deps, parser, detectors)
       .then(used => ({
         hits: {
-          [filename]: used.filter(dep => dep && dep !== '.' && dep !== '..'),
+          [filename]: used.filter(dep => dep && dep !== '.' && dep !== '..').reduce(unique, []),
         },
         used: used.filter(dep => dep && dep !== '.' && dep !== '..'),
       }), error => ({
--- a/src/index.js
+++ b/src/index.js
@@ -148,7 +148,7 @@ function checkFile(dir, filename, deps, parsers, detectors) {
     getDependencies(dir, filename, deps, parser, detectors)
       .then(used => ({
         hits: {
-          [filename]: used.filter(dep => dep && dep !== '.' && dep !== '..')
+          [filename]: used.filter(dep => dep && dep !== '.' && dep !== '..'),
         },
         used: used.filter(dep => dep && dep !== '.' && dep !== '..'),
       }), error => ({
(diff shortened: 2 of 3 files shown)


lijunle:
Two questions:

1. Why do you want to show the used dependencies?
2. I love the idea to show how each dependencies is used (with file name, row and column number *if available*). Not only for the `used` part, but also the `missing` part (which is added recently to show the dependencies are in used but not in package.json).

```js
{
  "react": ["/file1.js:3:3", "/file2.js:4:5"],
  "babel": ["/package.json"]
}
```

gtanner:
1.  I want to know the level of which someone is using a dependency which can help during upgrades of the version of that dependency or when I want to refactor out or replace a dependency.
2.  I love that idea too, the current matcher doesn't return that data at the moment (that I think), but the line number for the import would be great to have.  (would require re adding back in the duplicates in the list of usages.

src/cli.js
```diff
@@ -45,6 +45,7 @@ export default function cli(args, log, error, exit) {
     .describe('dev', 'Check on devDependecies')
     .describe('ignore-bin-package', 'Ignore package with bin entry')
     .describe('json', 'Output results to JSON')
+    .describe('used', 'Report on which dependencies are used')
```
lijunle: I think we could skip this parameter and always consider it as `true`

src/index.js
```diff
@@ -147,6 +147,9 @@ function checkFile(dir, filename, deps, parsers, detectors) {
   return targets.map(parser =>
     getDependencies(dir, filename, deps, parser, detectors)
       .then(used => ({
+        hits: {
```
lijunle: Rename this as `used` and remove the existing `used`.

src/index.js
```diff
@@ -234,6 +239,12 @@ export default function depcheck(rootDir, options, callback) {
       dependencies: minus(deps, result.used),
```
lijunle: Build up `dependencies` and `devDependencies` from your new `used` object.

src/index.js
```diff
@@ -234,6 +239,12 @@ export default function depcheck(rootDir, options, callback) {
...
       devDependencies: minus(devDeps, result.used),
       missing: minus(result.used, allDeps).filter(dep => builtInModules.indexOf(dep) === -1),
+      used: Object.keys(result.hits).reduce((obj, current) => {
```
lijunle: Abstract the complex code as a single function please.

src/index.js
```diff
@@ -234,6 +239,12 @@ export default function depcheck(rootDir, options, callback) {
...
+        return result.hits[current].reduce((o, d) => {
```
lijunle: I think ESLint should fail your `o` and `d` variable names. I don't know why CI still pass. :(

src/utils/output.js
```diff
@@ -17,17 +17,24 @@ function prettify(caption, deps) {
...
   return new Promise(resolve => {
-    if (json) {
-      log(JSON.stringify(result, replacer));
+    if (argv.json) {
+      const clone = Object.assign({}, result);
+      if (!argv.used) delete clone.used;
+      log(JSON.stringify(clone, replacer));
     } else if (noIssue(result)) {
-      log('No depcheck issue');
+      const unused = ['No depcheck issue'];
```
lijunle: I am not sure if we need to show ***Used dependencies* in this case. Maybe user with the common scenario are not going to see a long list. However, if you are going to show the list, I think it could better to remove the **No depcheck issue** caption.

src/utils/output.js
```diff
@@ -17,17 +17,24 @@ function prettify(caption, deps) {
...
+      const used = prettify('Used Dependencies', Object.keys(result.used));
+      const content = unused.concat(argv.used ? used : []).join('\n');
+
+      log(content);
     } else {
       const deps = prettify('Unused dependencies', result.dependencies);
       const devDeps = prettify('Unused devDependencies', result.devDependencies);
       const missing = prettify('Missing dependencies', result.missing);
-      const content = deps.concat(devDeps, missing).join('\n');
+      const used = prettify('Used Dependencies', Object.keys(result.used));
```
lijunle: Use "Used **d**ependencies"

lijunle:
Hi, I added some comments on your review. Could you please take a look?

A basic idea is, I am not sure if we need to show **Used dependencies** in the case everything is OK. Maybe user with the common scenario are not going to see a long list. However, if you are going to show the list, I think it could better to remove the **No depcheck issue** caption.

lijunle:
@gtanner I have taken care of this feature. The CLI output is not affected. However, the API and the JSON view (--json) will output the using dependencies list.

In regards to the reference row and column number, it has not been implemented. It needs more discussion.

gtanner:
Thanks!

PR #98, opened 2016-01-25
Account for io being async

gtanner:
If there are a lot of results, there is a chance of exiting this process before it is finished writing to the console.

Wait for stdout to be finished before exiting.

### DIFF
--- a/bin/depcheck
+++ b/bin/depcheck
@@ -6,4 +6,9 @@ require('../dist/cli')(
   process.argv.slice(2),
   console.log,
   console.error,
-  process.exit);
+  function (code) {
+    // ensure all data is flushed before terminating the process
+    process.stdout.on('finish', function () {
+      process.exit(code);
+    });
+  });


lijunle:
Is this serious? There is only one call for `console.log` or `console.error`. I think these two operations are synchronous and block the IO. Then, is it no need to wait the finish?

gtanner:
Yeah ;)  `console.error` and `console.warn` are sync, but `console.log` is async, thus needing to wait for stdout to flush.

I was getting half generated JSON output on some larger objects

gtanner:
It all depends where the "output" is going: https://nodejs.org/dist/latest-v4.x/docs/api/console.html#console_asynchronous_vs_synchronous_consoles

gtanner:
But in the case of someone using `child_process.exec` to run this, the output is always a stream, which forces stdout to be async.

lijunle:
Great information! Merged! Thanks very much! :beers:

Issue #107, opened 2016-03-08
Allow to define patterns against absolute file path


tamasmagedli:
First of all it's a great tool and I wanted to integrate it into my project, though the use case of mine is a little bit different how you prefer to use it. Acutally I wanted to parse only a specific file list within the root directory so ignoring dirs doesn't help.

Reading the codebase I realised that I have another option, namely I can define parsers for patterns which is a cool a feature though it cannot be used for my use case, at least not for now.

Let's assume that we have two JS files within the `src` folder, and I want to use this pattern:
`src/*.js`

Currently it does not work as you get the basename from the absolute path of the file.
line 124 in check.js: 
`const basename = path.basename(filename);`
This prevents to use a prefix pattern for the absolute path like mentioned before.

Can you please omit the basename call and pass the filename to the minimatch?

Thank you for fixing this.

lijunle:
I think this is a good change and all existing globs can be backward compatible because the `*` in `*.js` will match the dirname part.

Could you please send a pull request to implement it?

lijunle:
Besides, it could be great to add one more test case to support the feature. :D

tamasmagedli:
Yeah, I'll do that with new tests just wanted to know your opinion on this prior the changes.

lijunle:
Closed by #108.

PR #108, opened 2016-03-08
Support parser patterns based on file paths

tamasmagedli:
Original issue: https://github.com/depcheck/depcheck/issues/107


tamasmagedli:
Yeah, that's an ugly bug.

You're right, unfortunately I created the dist folder in my first commit which was obviously ignored by git :)

Issue #119, opened 2016-03-21
Support for commitizen modules


LinusU:
[Commitizen](https://github.com/commitizen/cz-cli) is a tool for enforcing style of git commit messages, it's quite useful. However, `npm-check` is giving me a hard time about unused dependencies.

```text
test-cz       😕  NOTUSED?  Possibly never referenced in the code.
                            npm uninstall --save-dev test-cz to remove.
```

My package.json looks like this:

```json
{
  "devDependencies": {
    "test-cz": "^0.1.0"
  },
  "config": {
    "commitizen": {
      "path": "./node_modules/test-cz"
    }
  }
```

LinusU:
btw. I don't know why commitizen thought it was a good idea to use `./node_modules/test-cz` instead of just `test-cz` (it internally uses `require.resolve` so both should work) but that is how it does when running `commitizen init test-tz --save-dev`...

lijunle:
It looks an excellent tool. Do you want to help to support it?

BTW, both `./node_modules/test-cz` and `test-cz` are OK. We can use the same logic to resolve the path.

jimthedev:
Commitizen author here. @LinusU you can use either syntax in commitizen too. :) We need to update that in the docs, it is a historical artifact but was fixed recently in a PR that made it possible to use either. The main reason we offer it is if you want to write a custom adapter then we want you to be able to just point to a path. <3

LinusU:
@lijunle I have this bookmarked and I would love to send a pr, I can't give any guarantees on when though but hopefully I'll get some time this weekend.

@jimthedev Cool, let's support both here as well. As a side note, would it be possible to update `commitizen init` to put in `test-cz` instead of `./node_modules/test-cz`? I think it looks cleaner...

LinusU:
Pull request submitted :raised_hands:

lijunle:
Resolved by #122, thanks to @LinusU :D

PR #122, opened 2016-03-24
Add commitizen support

LinusU:
See #119

### DIFF
--- a/src/special/commitizen.js
+++ b/src/special/commitizen.js
@@ -1,0 +1,32 @@
+import path from 'path';
+import requirePackageName from 'require-package-name';
+
+export default function parseCommitizen(content, filePath, deps, rootDir) {
+  const packageJsonPath = path.resolve(rootDir, 'package.json');
+  const resolvedFilePath = path.resolve(filePath);
+
+  if (resolvedFilePath === packageJsonPath) {
+    const metadata = JSON.parse(content);
+
+    if (metadata.config && metadata.config.commitizen && metadata.config.commitizen.path) {
+      const commitizenPath = metadata.config.commitizen.path;
+
+      if (!commitizenPath.startsWith('.')) {
+        return [requirePackageName(commitizenPath)];
+      }
+
+      const normalizedPath = path.normalize(commitizenPath).replace(/\\/g, '/');
+
+      if (!normalizedPath.startsWith('node_modules')) {
+        // The path is not refering to a file in another module
+        return [];
+      }
+
+      const packagePath = normalizedPath.substring('node_modules/'.length);
+
+      return [requirePackageName(packagePath)];
+    }
+  }
+
+  return [];
+}
--- a/test/special/commitizen.js
+++ b/test/special/commitizen.js
@@ -1,0 +1,51 @@
+/* global describe, it */
+
+import 'should';
+import commitizenSpecialParser from '../../src/special/commitizen';
+
+describe('commitizen special parser', () => {
+  it('should ignore when it is not `package.json`', () => {
+    const result = commitizenSpecialParser('content', '/a/file', [], '/a');
+    result.should.deepEqual([]);
+  });
+
+  it('should ignore when path is missing', () => {
+    const metadata = {
+      config: {
+        commitizen: {},
+      },
+    };
+
+    const content = JSON.stringify(metadata);
+    const result = commitizenSpecialParser(content, '/a/package.json', [], '/a');
+    result.should.deepEqual([]);
+  });
+
+  it('should recognize the module used by commitizen (long style)', () => {
+    const metadata = {
+      config: {
+        commitizen: {
+          path: './node_modules/cz-test',
+        },
+      },
+    };
+
+    const content = JSON.stringify(metadata);
+    const result = commitizenSpecialParser(content, '/a/package.json', [], '/a');
+    result.should.deepEqual(['cz-test']);
+  });
+
+  it('should recognize the module used by commitizen (short style)', () => {
+    const metadata = {
+      config: {
+        commitizen: {
+          path: 'cz-test',
+        },
+      },
+    };
+
+    const content = JSON.stringify(metadata);
+    const result = commitizenSpecialParser(content, '/a/package.json', [], '/a');
+    result.should.deepEqual(['cz-test']);
+  });
+});


src/special/commitizen.js
```diff
@@ -0,0 +1,22 @@
+import path from 'path';
+
+export default function parseFerossStandard(content, filePath, deps, rootDir) {
```
lijunle: Change the function name `parseFerossStandard` to `parseCommitizen`.

LinusU: Ooops :) will fix

lijunle:
It seems [ESLint is broken](https://travis-ci.org/depcheck/depcheck/jobs/118274049#L917) again. :cry: 

Could you try run `npm run lint` offline?

LinusU:
I tried but it was broken locally as well :/

In addition, four unrelated tests are failing on my local machine:

```text
  4 failing

  1) depcheck command line should support Typescript syntax:

      AssertionError: expected Array [ 'react', 'ts-dep-1', 'ts-dep-2', 'unused-dep' ] to equal Array [ 'unused-dep' ] (at length, A has 4 and B has 1)
      + expected - actual

       [
      -  "react"
      -  "ts-dep-1"
      -  "ts-dep-2"
         "unused-dep"
       ]
      
      at Assertion.fail (node_modules/should/lib/assertion.js:92:17)
      at Assertion.Object.defineProperty.value (node_modules/should/lib/assertion.js:164:19)
      at cli.js:68:36
      at run (node_modules/babel-polyfill/node_modules/core-js/modules/es6.promise.js:89:22)
      at node_modules/babel-polyfill/node_modules/core-js/modules/es6.promise.js:102:28
      at flush (node_modules/babel-polyfill/node_modules/core-js/modules/_microtask.js:14:5)

  2) depcheck command line should support SASS/SCSS syntax:

      AssertionError: expected Array [ 'sass-dep', 'scss-dep', 'unused-sass-dep' ] to equal Array [ 'unused-sass-dep' ] (at length, A has 3 and B has 1)
      + expected - actual

       [
      -  "sass-dep"
      -  "scss-dep"
         "unused-sass-dep"
       ]
      
      at Assertion.fail (node_modules/should/lib/assertion.js:92:17)
      at Assertion.Object.defineProperty.value (node_modules/should/lib/assertion.js:164:19)
      at cli.js:68:36
      at run (node_modules/babel-polyfill/node_modules/core-js/modules/es6.promise.js:89:22)
      at node_modules/babel-polyfill/node_modules/core-js/modules/es6.promise.js:102:28
      at flush (node_modules/babel-polyfill/node_modules/core-js/modules/_microtask.js:14:5)

  3) depcheck should support Typescript syntax:

      AssertionError: expected Array [ 'react', 'ts-dep-1', 'ts-dep-2', 'unused-dep' ] to equal Array [ 'unused-dep' ] (at length, A has 4 and B has 1)
      + expected - actual

       [
      -  "react"
      -  "ts-dep-1"
      -  "ts-dep-2"
         "unused-dep"
       ]
      
      at Assertion.fail (node_modules/should/lib/assertion.js:92:17)
      at Assertion.Object.defineProperty.value (node_modules/should/lib/assertion.js:164:19)
      at index.js:37:36
      at run (node_modules/babel-polyfill/node_modules/core-js/modules/es6.promise.js:89:22)
      at node_modules/babel-polyfill/node_modules/core-js/modules/es6.promise.js:102:28
      at flush (node_modules/babel-polyfill/node_modules/core-js/modules/_microtask.js:14:5)

  4) depcheck should support SASS/SCSS syntax:

      AssertionError: expected Array [ 'sass-dep', 'scss-dep', 'unused-sass-dep' ] to equal Array [ 'unused-sass-dep' ] (at length, A has 3 and B has 1)
      + expected - actual

       [
      -  "sass-dep"
      -  "scss-dep"
         "unused-sass-dep"
       ]
      
      at Assertion.fail (node_modules/should/lib/assertion.js:92:17)
      at Assertion.Object.defineProperty.value (node_modules/should/lib/assertion.js:164:19)
      at index.js:37:36
      at run (node_modules/babel-polyfill/node_modules/core-js/modules/es6.promise.js:89:22)
      at node_modules/babel-polyfill/node_modules/core-js/modules/es6.promise.js:102:28
      at flush (node_modules/babel-polyfill/node_modules/core-js/modules/_microtask.js:14:5)
```

src/special/commitizen.js
```diff
@@ -0,0 +1,22 @@
...
+
+export default function parseCommitizen(content, filePath, deps, rootDir) {
+  const packageJsonPath = path.resolve(rootDir, 'package.json');
+  const resolvedFilePath = path.resolve(filePath);
+
+  if (resolvedFilePath === packageJsonPath) {
+    const metadata = JSON.parse(content);
+
+    if (metadata.config && metadata.config.commitizen && metadata.config.commitizen.path) {
+      const {path} = metadata.config.commitizen;
```
lijunle: I think you will get ESLint warnings here, because `path` is already defined from `import path from 'path'`. Could you please wait and rebase on my ESLint fix? We should ensure ESLint from CI build.

LinusU: Ahh, true.

Of course, no stress :+1:

src/special/commitizen.js
```diff
@@ -0,0 +1,22 @@
...
+  const packageJsonPath = path.resolve(rootDir, 'package.json');
+  const resolvedFilePath = path.resolve(filePath);
+
+  if (resolvedFilePath === packageJsonPath) {
+    const metadata = JSON.parse(content);
+
+    if (metadata.config && metadata.config.commitizen && metadata.config.commitizen.path) {
+      const commitizenPath = metadata.config.commitizen.path;
+
+      if (commitizenPath.startsWith('./node_modules/')) {
```
lijunle: The separator (`/`) is OS specific and bad. 

You could do this:

```js
const resolvedPluginPath = require.resolve(commitizenPath); // full path
const relatedPluginPath = path.relative(rootDir, resolvedPluginPath); // *maybe* node_module relative path
if (relatedPluginPath.startsWith('node_module')) {
  const normalizedPluginPath = relativePluginPath.substring('node_module/').replace(/\\/g, '/');
  const pluginName = requirePackageName(normalizedPluginPath); // require-package-name will handle the scoped package
  // finally, we get the plugin name here...
}
```

src/special/commitizen.js
```diff
@@ -0,0 +1,22 @@
...
+        return [commitizenPath.substring(15)];
+      } else if (!commitizenPath.startsWith('./')) {
```
lijunle: Besides these two cases, the user may write full OS path inside the box too. My above code snippet will handle that.

src/special/commitizen.js
```diff
@@ -0,0 +1,22 @@
+import path from 'path';
+
+export default function parseCommitizen(content, filePath, deps, rootDir) {
+  const packageJsonPath = path.resolve(rootDir, 'package.json');
+  const resolvedFilePath = path.resolve(filePath);
+
+  if (resolvedFilePath === packageJsonPath) {
```
lijunle: One more thing, you should check the `commitizen` package itself is inside the `deps` array at the first beginning.

test/special/commitizen.js
```diff
@@ -0,0 +1,51 @@
...
+
+  it('should ignore when path is missing', () => {
+    const metadata = {
+      config: {
+        commitizen: {},
+      },
+    };
+
+    const content = JSON.stringify(metadata);
+    const result = commitizenSpecialParser(content, '/a/package.json', [], '/a');
```
lijunle: Add the `commitizen` package to the deps array.

lijunle:
@LinusU Thanks for sending PR, please resolve the comments. I am resolving the ESLint failure. Once that is done, please rebase your branch to newest one and squash your commits. Great thanks! :smile_cat:

src/special/commitizen.js
LinusU: Actually, you usually don't have commitizen as a dependency. Instead it's installed globally.

lijunle: Oh, you are right. That keep this current behavior. (Another special may has this kind of issue, however...)

lijunle:
@LinusU Done. Could you please rebase your changes on the new master?

Besides, with the default `npm i`, I think you are getting some default failing test cases. Those will be passed with peer dependencies installed. I am resolving that problem. However, that will not block your merge. 

Thanks! :D

LinusU:
Absolutely, I'm on it :+1: 

btw, I could install them locally, which modules is it?

lijunle:
```sh
npm install typescript node-sass
```

LinusU:
Great, now the tests works!

However, I'm having some problems with the code snippet you suggested. `require.resolve` will only work if the packet is actually installed, is that really what we want?

Maybe I'm missing something...

> Besides these two cases, the user may write full OS path inside the box too. My above code snippet will handle that.

But if they write a full path, it's probably not a module? Maybe it would be best to just only support the plain syntax (only package name). We don't want to parse all possible values for the config, just find out wether or not it's a module that is used.

Thoughts?

LinusU:
For reference, when using `require.resolve`, this error is thrown:

```text
Error: Cannot find module './node_modules/cz-test'
```

Also, won't `require.resolve` be relative to the file I'm in (`src/special/commitizen.js`) and then always fail when it starts with `./`?

lijunle:
> However, I'm having some problems with the code snippet you suggested. require.resolve will only work if the packet is actually installed, is that really what we want?

Yes, we did the same thing for SASS parser. Production code is [here](https://github.com/depcheck/depcheck/blob/master/src/parser/sass.js#L14). A simple test code is created [here](https://github.com/depcheck/depcheck/tree/master/test/fake_modules/sass).

> Error: Cannot find module './node_modules/cz-test'

Just create a simple package (with `package.json` and `index.js`) to `special/node_module` folder.

LinusU:
> Production code is here.

I don't see `require.resolve` being used here?

LinusU:
`require-package-name` looks awesome though :raised_hands:

lijunle:
> I don't see require.resolve being used here?

Sorry, you should not use `require.resolve`. I think two cases for *module name* and *relative path* (start with '.') is fine.

I am not sure how to handle `node_module/plugin/index.js`. May it is not supported.

lijunle:
If the value starts with '.', `path.resolve(rootpath, value)`, then relative, normalize, require package name.

LinusU:
Got it now :)

`node_module/plugin/index.js` is supported, as well as referring to any other file within a module.

I think I have a awesome solution, submitting it now :raised_hands:

lijunle:
Merged, thanks! :+1:

LinusU:
Thank you :tada:

PR #125, opened 2016-03-24
Mark the peer dependencies as dev dependencies too.

lijunle:
/cc @LinusU

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,6 @@ node_js:
   - 5
 
 script:
-  - npm install typescript node-sass # peer dependencies
   - npm run depcheck
   - npm run depcheck-json
   - npm run depcheck-web
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -13,7 +13,6 @@ install:
 test_script:
   - node --version
   - npm --version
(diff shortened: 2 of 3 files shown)


LinusU:
Isn't the accepted practice that you include your dependencies peer-dependencies as your own dependencies? That is, just adding them as dependencies...

Also, I had a warning about `eslint-plugin-react` being an unmet peer dependency as well. Maybe that should go on the list also...

lijunle:
We do not use `eslint-plugin-react` in our code. That one is fine to let it leave.

Not everybody use typescript and SASS parser, to whom use them, they will install typescript and SASS parser at project level. That is exactly how peer dependency works.

I have ideas to resolve these failing tests. The idea is to mark them as dev dependencies and peer dependencies.

LinusU:
Ignore my stupid comment, I thought that it was peer dependencies to one of our dependencies :laughing:

lijunle:
Actually, they are. We depends on their parse function to parse `.ts` (Typescirpt) files, `.sass` and `.scss` (SASS) files. However, I don't want to install everything then the user only use plain JavaScript.

PR #127, opened 2016-03-25
Support gulp plugins loaded by gulp-load-plugins.

### LINKED ISSUE #18: False Positives with gulp-load-plugins && gulp-jshint
Hello,

Was looking around at how to determine un-used npm packages and found your project. 

When I ran it the following report was generated. However, due to the way I'm loading gulp plugins (using `gulp-load-plugins`, my guess is this tool is not picking up anything that is dynamically loaded. Similarly, I'm asking it to use the `jshint-stylish` reporter for my jshint, which I presume is another dynamically loaded thing...

```
Unused Dependencies
* gulp-debug

Unused devDependencies
* gulp-markdown
* gulp-mocha
* gulp-plumber
* gulp-jshint
* gulp-jscs
* jshint-stylish
* gulp-istanbul
```

I can't say I have any bright ideas on how you can solve these problems, but in case you hadn't heard of them, they might be interesting ones to look into.

lijunle:
Related to #18 

- Use `babel-traverse` to walk through the AST to figure out which packages are loaded by gulp-load-plugins.
- Add simple test cases to support.
- Only `gulp-*` and `gulp.*` are supported.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
   "license": "MIT",
   "readmeFilename": "README.md",
   "dependencies": {
+    "babel-traverse": "^6.7.3",
     "babylon": "^6.1.21",
     "builtin-modules": "^1.1.1",
     "deprecate": "^0.1.0",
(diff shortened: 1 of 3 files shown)


lijunle:
Still not cover:

- Package: `gulp-sourcemap`
- Code: `$.sourcemaps.init()`

PR #141, opened 2016-06-01
Feat/refinement nodejs in ci

iamstarkov:
for some reason, tests on my machine are not passing, so i took a look at nodejs versions used for testing in CI and found a room for improvement:

* add v6, cause its latest and will be lts in the future
* remove v3, because its not one of the latest and not LTS as well
* add v0.10, because its LTS and imho every tool should work there to support enterprise

About the latest thing: from my experience only thing to do here is to polyfill promises and latest nodejs core APIs — i can help here.

what do you think?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ node_js:
   - 3
   - 4
   - 5
+  - 6
 
 script:
   - npm run depcheck
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ sudo: false
 language: node_js
 
 node_js:
+  - 0.10
   - 0.12
   - 4
   - 5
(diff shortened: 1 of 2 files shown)


lijunle:
Hey, @iamstarkov 

Thanks for your pull request. I am OK to have 0.10 support, if the promise is the only gap, sound like good to support it. I am afraid some of ES6 API needs three Babel runtime package.

Btw, if you are going to do so, could you please update readme and engine version in package.json.

iamstarkov:
i gave up =(

lijunle:
Hey, @iamstarkov , I checked the node.js support [schedule](https://github.com/nodejs/lts), the v0.10 is supported till Oct 2016. Now, it is June. I am fine to drop its support now.

I ever got few questions about *Promise not defined* (v0.10 issue). They are OK to upgrade their node to a newer version.

Do you have more concerns here?

PR #152, opened 2016-06-19
Enable AppVeyor cache to increase build speed.

lijunle:
Fix the typo in README.


lijunle:
AppVeyor is not the bottleneck of the CI build now. No need to do special hack on it.

PR #165, opened 2016-09-05
Update out of date devDependencies

domharrington:
https://david-dm.org/depcheck/depcheck#info=devDependencies


lijunle:
@domharrington Sorry that, I stop the depcheck web service. You could simply remove the command [`depcheck-web`](https://github.com/depcheck/depcheck/blob/master/.travis.yml#L13) from `.travis.yml`.

PR #166, opened 2016-09-21
ignore TypeScript `@types` packages by default

### LINKED ISSUE #163: Add support for @types declaration packages
Typescript v2 is moving to support a new approach to adding typings to a project.

Type definitions will be included in packages contained within the @types scope (e.g. @types/mypackage), which can simply be installed via npm (just like other packages).

More information can be found at:
https://blogs.msdn.microsoft.com/typescript/2016/06/15/the-future-of-declaration-files/

Currently any of these @types packages that are referenced in package.json, are incorrectly recognised as unused dependencies.

FredKSchott:
Fixes #163

The "@types/" scope is reserved for type declarations used in TypeScript projects. They are only used during build and are never referenced in the application.

This change ignores them by default so that depcheck doesn't fail for TypeScript projects by default. In the future, we could add additional logic to check each type package and detect if it's sibling-package (that it declares types for) exists as well.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -40,8 +40,11 @@ export default function depcheck(rootDir, options, callback) {
 
   const withoutDev = getOption('withoutDev');
   const ignoreBinPackage = getOption('ignoreBinPackage');
-  const ignoreMatches = getOption('ignoreMatches');
   const ignoreDirs = lodash.union(defaultOptions.ignoreDirs, options.ignoreDirs);
+  // The "@types/" scope is reserved for type declarations used in TypeScript projects. They are only used
+  // during build and are never referenced in the application. We ignore them by default so that depcheck
+  // doesn't fail for TypeScript projects by default.
+  const ignoreMatches = getOption('ignoreMatches') || ['@types/*'];
 
   const detectors = getOption('detectors');
   const parsers = lodash(getOption('parsers'))
--- a/src/index.js
+++ b/src/index.js
@@ -41,10 +41,12 @@ export default function depcheck(rootDir, options, callback) {
   const withoutDev = getOption('withoutDev');
   const ignoreBinPackage = getOption('ignoreBinPackage');
   const ignoreDirs = lodash.union(defaultOptions.ignoreDirs, options.ignoreDirs);
-  // The "@types/" scope is reserved for type declarations used in TypeScript projects. They are only used
-  // during build and are never referenced in the application. We ignore them by default so that depcheck
-  // doesn't fail for TypeScript projects by default.
-  const ignoreMatches = getOption('ignoreMatches') || ['@types/*'];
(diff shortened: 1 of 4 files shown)


lijunle:
Hi, @FredKSchott. Thank you for taking time on this issue. We should `.concat` instead of `||`. Btw, could you please add a test for it?

FredKSchott:
@lijunle thanks for taking a look, I've pushed a new set of changes & added a test

src/index.js
```diff
@@ -40,8 +40,13 @@ export default function depcheck(rootDir, options, callback) {
...
   const withoutDev = getOption('withoutDev');
   const ignoreBinPackage = getOption('ignoreBinPackage');
-  const ignoreMatches = getOption('ignoreMatches');
   const ignoreDirs = lodash.union(defaultOptions.ignoreDirs, options.ignoreDirs);
+  const ignoreMatches = getOption('ignoreMatches');
+
+  // The "@types/" scope is reserved for type declarations used in TypeScript projects. They are
+  // only used during build and are never referenced in the application. We always ignore them
+  // so that depcheck doesn't fail for TypeScript projects by default.
+  ignoreMatches.push('@types/*');
```
lijunle: Please use `.concat` instead of `.push`.

```
const ignoreMatches = ['@types/*'].concat(getOption('ignoreMatches'))
```

test/index.js
```diff
@@ -54,6 +54,18 @@ describe('depcheck', () => {
       error.should.be.instanceof(SyntaxError);
     }));
 
+  it('should ignore typescript `@types` packages', () =>
```
lijunle: This should goes to spec.js file.

test/index.js
```diff
@@ -54,6 +54,18 @@ describe('depcheck', () => {
...
+    check('ignore_npm_@types', {}).then(unused => {
+      unused.should.deepEqual({
+        dependencies: [ 'optimist' ],
```
lijunle: please run  `npm run lint` to follow code style.

lijunle:
@FredKSchott Are you still working on this? I think we could better to make the `@types` support as a _special detector_ instead of completely ignore them. How do you think about this?

FredKSchott:
Sorry I completely lost track of this PR :) I'd be happy to respond to these comments and help get this through, but I don't know if I'll have the time to work on a more complex detector anytime soon. Would you still like me to work to get this through?

lijunle:
Sure, if you have time! Thank you for contributing! 👍

rjmunro:
Ideally @types packages should not be ignored. They should check for usage of whatever modules they are the types for. E.g. `@types/react` is only relevant if react is installed & used.

`@types/node` is a special case - this is needed for all nodeJS projects.

FredKSchott:
I believe this has been fixed

PR #193, opened 2016-12-11
Fix handling of scoped eslint plugins

corbinu:
I ran into some issues of depcheck showing my scoped plugin which contains my personal eslint config as both being unused and missing dependancies. 

It appears to be related to that the config is stored in a scoped plugin. This is the fix.

### DIFF
--- a/src/special/eslint.js
+++ b/src/special/eslint.js
@@ -45,6 +45,10 @@ function isEslintConfigARelativePath(specifier) {
   return lodash.startsWith(specifier, './') || lodash.startsWith(specifier, '../');
 }
 
+function isEslintPluginAScopedModule(specifier) {
+  return lodash.startsWith(specifier, 'plugin:@');
+}
+
 function isEslintConfigAScopedModule(specifier) {
   return lodash.startsWith(specifier, '@');
 }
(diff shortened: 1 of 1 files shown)


src/special/eslint.js
```diff
@@ -61,6 +65,15 @@ function resolvePresetPackage(preset, rootDir) {
   if (isEslintConfigARelativePath(preset)) {
```
corbinu: Sure will do

lijunle:
@corbinu 

Before you start fixing the comments. Could you please consider this suggestion? It is more important than fixing any comment!

> Thank you for contributing this project. I looked into [the current ESLint resolve logic](https://github.com/eslint/eslint/blob/da79f17bf24de631bb24ef422c8dada7261645c4/lib/config/config-file.js#L461). I find it is too complicated to shadow it here.
> My suggestion is to reference the ESLint package and logic directly. How do you think about this?

corbinu:
@lijunle I will take a look but wanted to make the other changes first ... as once that is done we can try simply replacing resolvePresetPackage(preset, rootDir) and see if it works. If it still passes then is great. Otherwise can move forward.

lijunle:
@corbinu In such case, we could better to have some test cases. Here are the cases I identified from the current ESLint implementation:

```php
plugin:@my-scope				=> @my-scoped/eslint-plugin
plugin:@my-scope/eslint-plugin			=> @my-scoped/eslint-plugin
plugin:@my-scope/my-plugin			=> @my-scoped/eslint-plugin-my-plugin
plugin:@my-scope/eslint-pluginmy-plugin		=> @my-scoped/eslint-plugin-my-plugin
plugin:eslint-plugin-my-plugin			=> eslint-plugin-my-plugin
plugin:my-plugin				=> eslint-plugin-my-plugin
```

For `eslint-config`, it basically has the same logic without the `plugin:` prefix.

corbinu:
Ok will take a look at the testing system when I have some time.

PR #196, opened 2017-01-09
add: support for require.resolve in JS

pwang2:
In cases of simple plugin registration, we use `require.resolve` which should also be processed by depcheck.


pwang2:
BTW, could you confirm the eslint rules?  It is asking for  trailing comma explicitly.
49:51  error  Missing trailing comma  comma-dangle

PR #204, opened 2017-02-16
A new detector to discover express view engine

aleung:
The PR implements a new detector `expressViewEngine`.

I don't add this detector to default options because I'm not sure if it should be enable by default.

The current test tool doesn't support giving `detectors` in `options` in `spec.js`. I have to modify it a bit. A tricky thing is that a naming rule is required for detectors to make cli test works: function name must be the same as its public name defined in `component.json`.


aleung:
@lijunle Could you review this PR?

PR #205, opened 2017-02-21
Add support for import() expressions

haggholm:
Add support for lazy `import()` calls as used by `System.import()` and lazy Webpack chunks.

Modelled on the `require.resolve()` PR (https://github.com/depcheck/depcheck/pull/196). Not too familiar with the test framework, but copying tests, it seems to work (and mutating tests does properly break…).

Perhaps this should be optional/disabled by default, but the options (esp. w.r.t. the tests) are a bit opaque to me: feedback extremely welcome.


test/fake_modules/system_import_missing/index.js
```diff
@@ -0,0 +1 @@
+System.import('anyone');
```
lijunle: I suppose the `System.import` has been deprecated and we don't need to support it.

test/fake_modules/import_function/index.js
```diff
@@ -0,0 +1 @@
+import('optimist');
```
lijunle: Could you please try the webpack named chunk feature (aka, the magic comment):

```js
import(/* webpackChunkName: 'chunk' */ 'chunk')
```

haggholm:
I guess I was about as quick to reply as you were! I’m not 100% what you mean; do these changes reflect what you wanted?

lijunle:
@haggholm Webpack's `import()` has a feature named [*magic comment*](https://medium.com/webpack/webpack-3-official-release-15fd2dd8f07b#ea65) to specify the chunk name with the following syntax:

```js
import(/* webpackChunkName: "my-chunk-name" */ 'module');
```

We should check if we are supporting it.

haggholm:
Right. I added a test that verifies that it still detects the import­ -- I think that’s what you meant? I also cleaned up some cruft I’d accidentally introduced.

mnkhouri:
Thank you for the contribution, @haggholm !  🎉

PR #223, opened 2017-09-20
Depcheck will detect dependency use in package.json scripts

### LINKED ISSUE #222: Consider dependencies used when used in scripts
In my `package.json` I use ts-node to quickly run my program while developing, but depcheck says ts-node isn't used at all.

```json
  "scripts": {
    "start": "node --require ts-node/register src/index.ts"
  },
```

```
> depcheck
Unused devDependencies
* ts-node

```

Since I am using ts-node in an npm script, it shouldn't be caught by depcheck as an unused dependency.

Ionaru:
* Depcheck will detect dependency use in package.json scripts.
* Added Jetbrains IDE files to .gitignore.
* Fixes #222

I don't know how to use the 'special' parsers with tests in spec.js, if someone could help me with that I'd be very grateful. :smile:

### DIFF
--- a/src/special/package.js
+++ b/src/special/package.js
@@ -1,0 +1,17 @@
+export default function parsePackageJson(content, filepath, deps) {
+
+  // Mark dependencies as "used" if they are used in a script in package.json.
+  if (filepath.endsWith('/package.json')) {
+    // Grab just the scripts section.
+    const scriptsText = JSON.stringify(JSON.parse(content)['scripts']);
+    const depsInScripts = [];
+    for (const dependency of deps) {
+      if (scriptsText.indexOf(dependency) !== -1) {
+        depsInScripts.push(dependency);
+      }
+    }
+    return depsInScripts;
+  }
+
+  return [];
+}
--- a/src/special/package.js
+++ b/src/special/package.js
@@ -1,16 +1,29 @@
 export default function parsePackageJson(content, filepath, deps) {
   // Mark dependencies as "used" if they are used in a script in package.json.
-  if (filepath.endsWith('/package.json')) {
+  if (filepath.match(new RegExp('(\\/|\\\\)package.json$'))) {  // Escaping is fun!
     // Grab just the scripts section.
-    const scriptsText = JSON.stringify(JSON.parse(content).scripts);
-    if (scriptsText) {
-      const depsInScripts = [];
-      for (let i = 0; i < deps.length; i += 1) {
-        if (scriptsText.indexOf(deps[i]) !== -1) {
-          depsInScripts.push(deps[i]);
+    const scripts = JSON.parse(content).scripts;
+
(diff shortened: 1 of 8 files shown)


Ionaru:
@lijunle @rumpl Can someone please have a look at this request and issue #222?

src/special/package.js
```diff
@@ -0,0 +1,30 @@
+export default function parsePackageJson(content, filepath, deps) {
+  // Mark dependencies as "used" if they are used in a script in package.json.
+  if (filepath.match(new RegExp('(\\/|\\\\)package.json$'))) {  // Escaping is fun!
```
lijunle: Why not use [`path.basename`](https://api.nodejs.org/dist/latest-v6.x/docs/api/path.html#path_path_basename_path_ext)? That could avoid this interesting regexp.

src/special/package.js
```diff
@@ -0,0 +1,30 @@
...
+    // Grab just the scripts section.
+    const scripts = JSON.parse(content).scripts;
+
+    if (scripts) {
+      const scriptKeys = Object.keys(scripts);
```
lijunle: You could use [`lodash.values()`](https://lodash.com/docs/#values) to get values directly.

src/special/package.js
```diff
@@ -0,0 +1,30 @@
...
+        // Don't count script titles as dependency use.
+        let scriptsText = '';
+        for (let i = 0; i < scriptKeys.length; i += 1) {
+          scriptsText += `${scripts[scriptKeys[i]]} `;
+        }
+
+        if (scriptsText) {
+          const depsInScripts = [];
+          for (let i = 0; i < deps.length; i += 1) {
+            if (scriptsText.indexOf(deps[i]) !== -1) {
```
lijunle: This may not work like that. 😢 

The script contains the dependency string does not *really* means it is using that dependency. 

---

BTW, even if compare string is acceptable, it needs to [grab the `bin` section](https://github.com/depcheck/depcheck/blob/master/src/special/bin.js#L41) from `dep/package.json` for comparison.

LinusU: btw. the `bin` field can be both a string or an object. When it's a string, the name of the binary is the same as the name of the package. When it's an object, every key corresponds to the binary name.

lijunle: @LinusU Good catch! Could you please fix the string case in the `bin` special? Thank you!

Ionaru:
My bad, I was under the impression the `bin` special did something entirely different. 

Is it okay if I add `--require ${key}/register` and `--require ${key}` to the `getBinaryFeatures` function so it will detect ts-node and coffee-script dependencies when used by mocha?

lijunle:
@Ionaru That will be good. Thank you!

Ionaru:
Done in #227

PR #224, opened 2017-10-01
fix typo in README.md --ignores flag explanation

leopoldjoy:
add missing asterisks around --ignores flag arguments to make them glob expressions

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ All of the arguments are optional:
 
 `--json`: Output results in JSON. When not specified, depcheck outputs in human friendly format.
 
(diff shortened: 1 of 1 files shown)


lijunle:
Hi, @leopoldjoy 

Thank you for raising this up. I think your concern is, the example does not show the glob example. Is that right?

For demonstration reason, it could better to have one with and one without glob. How about this - `--ignores=eslint,babel-*`?

leopoldjoy:
@lijunle Exactly, I felt it would be useful to show the glob example. I agree with you about showing both, I have updated this PR accordingly.

PR #226, opened 2017-10-10
feat: add webpack v2 support

jwhitmarsh:
I've updated the `special/webpack.js` file to parse new webpack v2 (and greater) config better, and added some tests to support the new functionality.

### DIFF
--- a/src/special/webpack.js
+++ b/src/special/webpack.js
@@ -51,18 +51,12 @@ function parseWebpack1(module, deps) {
   return [...loaders, ...preLoaders, ...postLoaders];
 }
 
-function mapRuleLoaders(module) {
-  return module.rules
-    .filter(rule => rule.loaders)
-    .map(rule => rule.loaders.map(loader => loader));
-}
-
 function mapRuleUse(module) {
   return module.rules
     // filter use or loader because 'loader' is a shortcut to 'use'
     .filter(rule => rule.use || rule.loader)
     // return coerced array, using the relevant key
-    .map(rule => [].concat(rule[rule.use ? 'use' : 'loader']));
+    .map(rule => [].concat(rule.use || rule.loader));
 }
 
 function parseWebpack2(module, deps) {
@@ -70,7 +64,7 @@ function parseWebpack2(module, deps) {
     return [];
   }
 
-  const mappedLoaders = mapRuleLoaders(module);
+  const mappedLoaders = module.rules.filter(rule => rule.loaders);
   const mappedUses = mapRuleUse(module);
   const loaders = getLoaders(deps, lodash.flatten([...mappedLoaders, ...mappedUses]));
   return loaders;
--- a/src/special/webpack.js
+++ b/src/special/webpack.js
@@ -42,8 +42,7 @@ function getLoaders(deps, loaders) {
     .value();
 }
 
-function parseWebpack1(filepath, deps) {
-  const module = require(filepath).module || {}; // eslint-disable-line global-require
+function parseWebpack1(module, deps) {
   const loaders = getLoaders(deps, module.loaders);
   const preLoaders = getLoaders(deps, module.preLoaders);
   const postLoaders = getLoaders(deps, module.postLoaders);
@@ -53,10 +52,9 @@ function parseWebpack1(filepath, deps) {
 function mapRuleLoaders(module) {
   return module.rules
     .filter(rule => rule.loaders)
-    .map(rule => rule.loaders
-      .map(loader => ({
-        loader,
-      })));
+    .map(rule => rule.loaders.map(loader => ({
+      loader,
+    })));
 }
 
 function mapRuleUse(module) {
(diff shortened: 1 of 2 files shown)


src/special/webpack.js
```diff
@@ -39,14 +42,82 @@ function getLoaders(deps, loaders) {
...
+}
+
+function mapRuleUse(module) {
+  return module.rules
+    // filter use or loader (loader is a shortcut to use)
+    .filter(rule => rule.use || rule.loader)
+    .map((rule) => {
+      const key = rule.use ? 'use' : 'loader';
+
+      // map simple strings as { loader: string }
```
lijunle: Instead of having three `if` clauses, I could suggest to convert the first two cases to `Array`, then handle it with the third case logic.

src/special/webpack.js
```diff
@@ -39,14 +42,82 @@ function getLoaders(deps, loaders) {
...
+  const preLoaders = getLoaders(deps, module.preLoaders);
+  const postLoaders = getLoaders(deps, module.postLoaders);
+  return [...loaders, ...preLoaders, ...postLoaders];
+}
+
+function mapRuleLoaders(module) {
+  return module.rules
+    .filter(rule => rule.loaders)
+    .map(rule => rule.loaders
+      .map(loader => ({
```
lijunle: Could better to inline:

```
.map(rule => rule.loaders.map(loader => ({ loader })));
```

src/special/webpack.js
```diff
@@ -39,14 +42,82 @@ function getLoaders(deps, loaders) {
...
+}
+
+function parseWebpack2(filepath, deps) {
+  try {
+    const module = require(filepath).module || {}; // eslint-disable-line global-require
+    if (!module.rules) {
+      return [];
+    }
+
+    const mappedLoaders = [];
```
lijunle: Directly construct `mappedLoaders` with `[...mapRuleLoader(), ...mapRuleUse()]`.

src/special/webpack.js
```diff
@@ -39,14 +42,82 @@ function getLoaders(deps, loaders) {
...
+    mappedLoaders.push(...mapRuleLoaders(module));
+    mappedLoaders.push(...mapRuleUse(module));
+
+    return getLoaders(deps, lodash.flatten(mappedLoaders));
```
lijunle: Is this `lodash.flatten` necessary? I think you have good structure at this moment.

jwhitmarsh: It is - `mapRuleUse` can return an array of arrays because `use` or `loader` can be an array.  Eg: 
```js
module: {
      rules: [
        { test: /\.css$/, loader: [{ loader: 'style-loader' }] },
      ],
    },
```

Rather than test the type of each `loader` or `use` property it's easier to just flatten the whole lot.

jwhitmarsh: I've made the rest of the changes - thank you for your excellent feedback! I'll wait until we've worked out what's going on here ^ before resubmitting the PR.

src/special/webpack.js
```diff
@@ -53,10 +52,9 @@ function parseWebpack1(filepath, deps) {
 function mapRuleLoaders(module) {
   return module.rules
     .filter(rule => rule.loaders)
-    .map(rule => rule.loaders
-      .map(loader => ({
-        loader,
-      })));
+    .map(rule => rule.loaders.map(loader => ({
+      loader,
+    })));
```
jwhitmarsh: Sorry, I did change this but it looks like it's auto formatted back to this. I'll change it again.

jwhitmarsh:
Apologies - I forgot that pushing to the branch would auto trigger this update (I thought i'd have to manually re-submit the PR). 

Not sure why codecov is now failing?

src/special/webpack.js
```diff
@@ -1,12 +1,16 @@
+/* eslint-disable no-console */
```
lijunle: I don't think we need this disable rule anymore.

src/special/webpack.js
```diff
@@ -39,14 +42,58 @@ function getLoaders(deps, loaders) {
...
+      });
+    });
+}
+
+function parseWebpack2(module, deps) {
+  if (!module.rules) {
+    return [];
+  }
+
+  return getLoaders(deps, lodash.flatten([...mapRuleLoaders(module), ...mapRuleUse(module)]));
```
lijunle: Directly return a complicated computation expression is not a good idea. It will make it hard to debug (inspect the return value). Split it to some small steps, it could make debugging easier.

src/special/webpack.js
```diff
@@ -1,12 +1,16 @@
...
+
 import path from 'path';
 import lodash from 'lodash';
 
 const webpackConfigRegex = /webpack(\..+)?\.config\.(babel\.)?js/;
 const loaderTemplates = ['*-webpack-loader', '*-web-loader', '*-loader', '*'];
 
 function extractLoaders(item) {
-  if (item.loader) {
+  if (item.loader && typeof item.loader === 'string') {
```
lijunle: I am thinking about if we can check `item` is a string, then directly return it. If that is possible, we don't need to have `.map(loader => ({ lodaer }))`

src/special/webpack.js
```diff
@@ -39,14 +42,58 @@ function getLoaders(deps, loaders) {
...
+
+function mapRuleUse(module) {
+  return module.rules
+    // filter use or loader (loader is a shortcut to use)
+    .filter(rule => rule.use || rule.loader)
+    .map((rule) => {
+      const key = rule.use ? 'use' : 'loader';
+      const coercedArray = [].concat(rule[key]);
+
+      // if it's an array, apply the two rules above for each element
```
lijunle: The comment has been outdated.

src/special/webpack.js
```diff
@@ -39,14 +42,58 @@ function getLoaders(deps, loaders) {
...
+      return coercedArray.map((value) => {
+        if (typeof value === 'object') {
+          return value;
+        }
+
+        return {
+          loader: value,
```
lijunle: If we check `typeof item === 'string'` in `extractLoaders`, this could be simplify too.

lijunle:
Thank you @jwhitmarsh for updating the PR.

The codecov is failing because the coverage is [dropping](https://codecov.io/gh/depcheck/depcheck/compare/004663011d9e6be55970199c34bf9fdf1e1a7cc8...ea8036cdbf4676daae8df02eae031bc64841f32b/src/src/special/webpack.js?before=src/special/webpack.js). It looks like there are two cases not covered in your test case.

jwhitmarsh:
I've made the changes you suggested and updated the tests to the coverage is back up to 100% (for webpack.js). If everything is good with this, let me know if you'd like me to rebase & squash it into one commit.

src/special/webpack.js
```diff
@@ -39,14 +44,46 @@ function getLoaders(deps, loaders) {
...
+  const loaders = getLoaders(deps, module.loaders);
+  const preLoaders = getLoaders(deps, module.preLoaders);
+  const postLoaders = getLoaders(deps, module.postLoaders);
+  return [...loaders, ...preLoaders, ...postLoaders];
+}
+
+function mapRuleLoaders(module) {
+  return module.rules
+    .filter(rule => rule.loaders)
+    .map(rule => rule.loaders.map(loader => loader));
```
lijunle: Sorry, why do we still need `.map(loader => loader)`?

src/special/webpack.js
```diff
@@ -39,14 +44,46 @@ function getLoaders(deps, loaders) {
...
+}
+
+function mapRuleUse(module) {
+  return module.rules
+    // filter use or loader because 'loader' is a shortcut to 'use'
+    .filter(rule => rule.use || rule.loader)
+    // return coerced array, using the relevant key
+    .map(rule => [].concat(rule[rule.use ? 'use' : 'loader']));
```
lijunle:
```
[].concat(rule.use || rule.loader)
```

src/special/webpack.js
jwhitmarsh: Sorry @lijunle, that was dumb of me :(!

lijunle:
Thank you @ jwhitmarsh ! It is merged now. Do you want a new version for this?

jwhitmarsh:
@lijunle It's entirely up to you - i've got a few other PR's coming your way shortly, but not sure exactly how long they will take.

lijunle:
OK, I will publish it on this Saturday. If you have new ideas, please ping
me in this time slot.

2017年10月12日 16:15，"James" <notifications@github.com>写道：

mattfysh:
@lijunle was this published to npm? the version I've got from latest doesn't have the webpack2 support

PR #234, opened 2018-01-18
Fix scoped ESLint module check

estei:
Based on ESLint documentation here:
https://eslint.org/docs/developer-guide/shareable-configs#npm-scoped-modules

The old support for ESLint scoped modules had a few issues:

It did not support a scoped module named `eslint-config`
This lead to a false positive missing dependency of
`@scope/eslint-config`

It did not support the shorthand that ESLint allows of just giving
the scope like this: `@scope` which ESLint behind the scenes makes
into `@scope/eslint-config`

It supported a version that ESLint specifically does not support
looking like this: `@scope/name`
The implementation would prefix the name with `eslint-config-`
making it `@scope/eslint-config-name` which could potentially lead to
resolution errors.

### DIFF
--- a/src/special/eslint.js
+++ b/src/special/eslint.js
@@ -53,6 +53,10 @@ function isEslintConfigAFullyQualifiedModuleName(specifier) {
   return lodash.startsWith(specifier, 'eslint-config-');
 }
 
+function isEslintConfigModuleJustAScope(specifier) {
+  return !lodash.includes(specifier, '/');
+}
+
 function resolvePresetPackage(preset, rootDir) {
   // inspired from https://github.com/eslint/eslint/blob/5b4a94e26d0ef247fe222dacab5749805d9780dd/lib/config/config-file.js#L347
   if (isEslintConfigAnAbsolutePath(preset)) {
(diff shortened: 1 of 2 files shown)


src/special/eslint.js
```diff
@@ -62,13 +66,12 @@ function resolvePresetPackage(preset, rootDir) {
     return path.resolve(rootDir, preset);
   }
   if (isEslintConfigAScopedModule(preset)) {
-    const scope = preset.substring(0, preset.indexOf('/'));
-    const module = preset.substring(preset.indexOf('/') + 1);
-
-    if (isEslintConfigAFullyQualifiedModuleName(module)) {
```
lijunle: Please notice this:

> The module name can also be customized, just note that when using scoped modules it is not possible to omit the eslint-config- prefix. Doing so would result in package naming conflicts, and thus in resolution errors in most of cases. For example a package named @scope/eslint-config-myconfig vs @scope/my-config, since both are valid scoped package names, the configuration should be specified as:

```json
{
    "extends": "@scope/eslint-config-myconfig"
}
```

It means, we still need to check `isEslintConfigAFullyQualifiedModuleName` here.

test/special/eslint.js
```diff
@@ -97,21 +97,30 @@ const testCases = [
     expected: [],
   },
   {
-    name: 'handle config of scoped module',
```
lijunle: According to the doc, we still need to support this test case. Could you please add it back? (And polish the name is a plus. 😄 )

43081j:
Fixed by #305

rumpl:
This was resolved by another PR

PR #235, opened 2018-01-22
CLI: exit with -1 when there are missing dependencies

valeriangalliat:
I'm not sure if it's a bug or a feature, and I didn't see any issue about that yet, but since it's a simple fix I was like, might as well make a PR.

Currently depcheck doesn't exits with an error when there's only missing dependencies, and doesn't provide an option to do so:

```
$ depcheck
Missing dependencies
* moment
$ echo $?
0
```

This PR exits with -1 also when there are missing dependencies.

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -118,8 +118,7 @@ export default function cli(args, log, error, exit) {
       skipMissing: opt.argv.skipMissing,
     }))
     .then(result => print(result, log, opt.argv.json))
-    .then(({ dependencies: deps, devDependencies: devDeps }) =>
-      exit(opt.argv.json || (deps.length === 0 && devDeps.length) === 0 ? 0 : -1))
+    .then(result => exit((opt.argv.json || noIssue(result)) ? 0 : -1))
     .catch((errorMessage) => {
       error(errorMessage);
       exit(-1);


lijunle:
@valeriangalliat I am sorry that I miss this PR.

This seems a bug to me. Instead of hacking around the CLI logic, could you please help to check why `depcheck()` call is not returning `-1` when there is only missing deps?

valeriangalliat:
No worries, thanks for taking the time to review!

I'm not sure I understand your comment though, the `depcheck` function doesn't return an exit code itself, it just returns an object with different properties like `dependencies`, `devDependencies` and `unused`, and as far as I understand, it's up to the consuming code to decide what it wants to error on.

So here for the CLI, it's just about adding a check for the `unused` property in addition to the already existing check for `dependencies` and `devDependencies` which result in exiting with an error code.

Or if you prefer I can add a property `hasErrors` to the `depcheck` API and then use that in the CLI so the logic to determine if there's an error or not would live in the `depcheck` function?

src/cli.js
```diff
@@ -118,8 +118,13 @@ export default function cli(args, log, error, exit) {
       skipMissing: opt.argv.skipMissing,
     }))
     .then(result => print(result, log, opt.argv.json))
-    .then(({ dependencies: deps, devDependencies: devDeps }) =>
-      exit(opt.argv.json || (deps.length === 0 && devDeps.length) === 0 ? 0 : -1))
+    .then(({ dependencies: deps, devDependencies: devDeps, missing }) => {
+      const hasNoErrors = deps.length === 0 &&
```
lijunle: Could you please make the variable `hasError`? A negative expression seems a little bit...

src/cli.js
```diff
@@ -118,8 +118,13 @@ export default function cli(args, log, error, exit) {
...
+                          devDeps.length === 0 &&
+                          Object.keys(missing).length === 0;
```
lijunle: Could you please wrap this check with `()` and each check on each line?

```
const hasError = (
  deps.length !== 0 ||
  devDeps.length !== 0 ||
  Object.keys(missing).length !== 0
);
```

src/cli.js
```diff
@@ -118,8 +118,13 @@ export default function cli(args, log, error, exit) {
...
+
+      exit(opt.argv.json || hasNoErrors ? 0 : -1);
```
lijunle: Could you please help to wrap `opt.argv.json || hasNoErrors` with `(...)`?

lijunle:
@valeriangalliat Oh, I am sorry that, I have almost forgot how `depcheck()` call works. With the context, I think the change is making sense.

I have some minor change requests. Could you please take a look? Thank you for fixing this!

src/cli.js
valeriangalliat: I called it `hasErrors` since there can be multiple errors still

src/cli.js
valeriangalliat: I just updated that, I used `> 0` instead of `!== 0` because it feels more idiomatic to me, if that's fine with you :)

valeriangalliat:
Actually I just see you updated master since the time I made that PR, I could reuse the `noIssue` function, don't merge just yet!

lijunle:
Oh, sorry for my interruption.

valeriangalliat:
It should be good now, even cleaner!

src/cli.js
```diff
@@ -118,8 +118,9 @@ export default function cli(args, log, error, exit) {
       skipMissing: opt.argv.skipMissing,
     }))
     .then(result => print(result, log, opt.argv.json))
-    .then(({ dependencies: deps, devDependencies: devDeps }) =>
-      exit(opt.argv.json || (deps.length === 0 && devDeps.length) === 0 ? 0 : -1))
+    .then(result => {
```
lijunle: Hmm... This is breaking the linter.

valeriangalliat: Oops! Updated

Issue #239, opened 2018-01-27
False dependency when using ESLint vue plugin


Milad:
Usage of `eslint-plugin-vue` in `.eslintrc.js`

```javascript
module.exports = {
    'extends': [
        'plugin:vue/essential',
        'plugin:vue/recommended',
        'plugin:vue/strongly-recommended'
    ],
    parser: 'vue-eslint-parser',
    parserOptions: {
        ecmaFeatures: {jsx: true},
        ecmaVersion: 6,
        sourceType: 'module'
    },
    env: {
        browser: true,
        es6: true,
        // jquery: true,
        node: true
    },
    globals: {module: false},
    plugins: ['vue'],
    rules: {}
};
```
`depcheck --json` output:

```json
{
  "missing": {
    "eslint-config-plugin:vue": [
      ".eslintrc.js"
    ]
  },
}
```

aecz:
Same error here with this config:
```
{
  "env": {
    "browser": false,
    "es6": true,
    "node": true,
    "mocha": true
  },
  "parserOptions": {
    "ecmaVersion": 2017,
    "sourceType": "module"
  },
  "plugins": ["security"],
  "extends": ["eslint:recommended", "plugin:security/recommended"],
  "rules": {
    "no-console": "warn"
  }
}
```
Result:
```
$ npx depcheck
Missing dependencies
* eslint-config-plugin:security
```

aecz:
I found a workaround. We can use the relative/absolute path syntax in the extends array and add the specific rules to activated them.
```
  "extends": [
    "eslint:recommended",
    "./node_modules/eslint-plugin-security/rules/detect-new-buffer.js",
    "./node_modules/eslint-plugin-security/rules/detect-non-literal-fs-filename.js,"
    "./node_modules/eslint-plugin-security/rules/detect-object-injection.js"
  ],
  "rules": {
    "security/detect-new-buffer": "warn",
    "security/detect-non-literal-fs-filename": "warn",
    "security/detect-object-injection": "warn",
  }
```

NetOperatorWibby:
The year is half over and this is still an issue. Shame.

VincentLanglet:
@rumpl This issue is fixed, I just tested it
Fixed by https://github.com/depcheck/depcheck/pull/309

rumpl:
Go ahead yeah. Thanks!

PR #244, opened 2018-02-23
Add handling ESLint plugin syntax in extends property

aecz:
Fix issue #239


LinusU:
I've moved the code for the linters to a common file in order to support tslint as well. Would you mind moving the patch over to `src/util/linters.js` (just replace `eslint` with `${flavour}`)?

Otherwise I'll do it for you when I have time 👍

43081j:
Fixed by #309 i suspect

VincentLanglet:
> Superseded by #309 i suspect

Agreed, @rumpl This can be close

PR #250, opened 2018-04-03
Support for eslint's 'plugin:' prefixed specifiers

buddydvd:
ESLint supports loading configuration files from plugins. In the `eslintrc.*`'s `extends` property, you can specify a config name with a special `plugin:` prefix. Like loading config from package modules, you can specify a short name in lieu of a full module name. However, for `plugin:` prefixed names,  short names resolve to `eslint-plugin-${shortName}` instead of `eslint-config-${shortName}`. This PR takes cares of that use case.

Here's the link to ESLint's documentation on loading config from plugins:
https://eslint.org/docs/user-guide/configuring#using-the-configuration-from-a-plugin

### DIFF
--- a/src/special/eslint.js
+++ b/src/special/eslint.js
@@ -45,12 +45,16 @@ function isEslintConfigARelativePath(specifier) {
   return lodash.startsWith(specifier, './') || lodash.startsWith(specifier, '../');
 }
 
-function isEslintConfigAScopedModule(specifier) {
+function isEslintConfigFromAPlugin(specifier) {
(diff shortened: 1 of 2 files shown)


znarf:
@buddydvd Thank you for contributing this patch.

I also ran into this issue, it's much needed.

lijunle:
@buddydvd @znarf Do you want a new version for it?

buddydvd:
I'm not in a hurry either. A semver minor bump is appreciated.

PR #252, opened 2018-04-18
Fix ignore-bin-package default in ReadMe

ronkorving:
The ReadMe incorrectly states that `ignore-bin-package` defaults to `true`. This is not the case.


mnkhouri:
Thank you, @ronkorving !

PR #258, opened 2018-07-04
Enable the plugins for Typescript parser, too

cdagli:
I was receiving some `SyntaxError: Unexpected token` errors on some files. So I tested with the new @babel/parser and received the below error: 

```
 SyntaxError: This experimental syntax requires enabling the parser plugin: 'objectRestSpread'
```

After that I've added the plugins wildcard option just like you had in the ES7 parser. And failing files are started to pass.


mnkhouri:
Thank you @cdagli ! This will be included in version `0.6.10`

Issue #265, opened 2018-08-09
Next Release (0.6.10)


mnkhouri:
Hi all, I've just joined in to assist in maintaining this project -- excited to help out! I have noticed that there are a few "low hanging fruit" / easy PRs and issues that we can resolve for a release soon, so I'm creating this issue + milestone to keep track of them.

After the release is out, I'd like do a little housekeeping, such as: check-in a package-lock.json, add node 10 to CI, consider updating some packages.

cc @LinusU and @rumpl, as I see you've been active recently :) obviously I am very new to this project, so please don't hesitate to share feedback

rumpl:
Hi!

I'm soooo glad you are doing things, thanks man! :) I really don't have much time or energy to take care of this project but I love it still dearly.

Maybe we can even publish `1.0.0` and then celebrate?

LinusU:
> check-in a package-lock.json, add node 10 to CI, consider updating some packages.

Great 👍 

> Maybe we can even publish 1.0.0 and then celebrate?

I think that neither me nor @mnkhouri is added to Npm, so someone else would have to push the release to npm.

rumpl:
I can publish on npm, once I get to know you guys I'll add you on the npm org, I'm not that confortable adding people too quickly to a package that is installed +22k times per week. I'm sure you guys can understand.

But that being said, I really am glad someone still wants to work on this project!

mnkhouri:
@LinusU I think [CI is set up](https://github.com/depcheck/depcheck/blob/master/.travis.yml#L30) for automatic releases to npm, but I haven't tested it or confirmed it working. @rumpl do you know? Totally understand for direct npm access :)

Re 1.0.0: I love the idea of a `1.0.0` since it makes it easier for consumers to confidently upgrade, I'm personally of the opinion that it makes sense to go to 1.0.0 sooner rather than later. I noticed there's an existing issue for version 1.0.0 with several API changes that @lijunle was hoping to make: https://github.com/depcheck/depcheck/issues/131. Perhaps after getting this next release out as a `0.x.x`, we could re-evaluate those API changes and get a milestone together for 1.0.0

LinusU:
Sounds good 👍 

Happy to help!

LinusU:
> I think CI is set up for automatic releases to npm

That's super!

I guess the release process is something like this then:

```sh
npm version (major|minor|patch)
git push --follow-tags
```

LinusU:
> I think CI is set up for automatic releases to npm

[...]

PR #267, opened 2018-08-09
Bump nyc to 13.x

LinusU:
It's about to be promoted to stable 🙌  istanbuljs/nyc#902

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2868,14 +2868,12 @@
         "balanced-match": {
           "version": "1.0.0",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "brace-expansion": {
           "version": "1.1.11",
           "bundled": true,
           "dev": true,
-          "optional": true,
           "requires": {
             "balanced-match": "^1.0.0",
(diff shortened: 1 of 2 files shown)


LinusU:
Ah right, Node 4.x support is dropped.

Hmm, I think we should drop that as well and release a new major (0.7.0) since Node.js 4.x is EOL.

coreyfarrell:
This module is has `"instrument": false` so this is not actually running the updated instrumenter included by nyc@13, it's using the old version included by babel-plugin-istanbul@4.  Updating the babel plugin takes more effort as it is a babel 7 plugin.

LinusU:
@mnkhouri added that here: 3f94e0b 🙌

mnkhouri:
@LinusU LGTM!

Ok to wait on merging this until after https://github.com/depcheck/depcheck/issues/265 , since it is a breaking change?

LinusU:
yeah, sounds good!

nkbt:
I think it lacks package-lock changes...

LinusU:
Rebased on master

PR #271, opened 2018-08-10
Adds a lockfile

mnkhouri:
I prefer `npm` over `yarn`, but open to discussion :)


LinusU:
> Adds a lockfile 

Very small thing; in the future it could be nice to follow these guidelines for commit messages: https://chris.beams.io/posts/git-commit/. So it would be "Add a lockfile", but no biggie ☺️

mnkhouri:
> nice to follow these guidelines for commit messages

I'll keep in mind for future commits :) thanks for the tip

> Could we add a .gitattributes file with this line:

👍 good call, I'm traveling this weekend but I'll add a commit to this branch once I'm back in action

LinusU:
Hmmm, one drawback of treating the package-lock.json as a binary file is that you cant `git add -p` or `git diff` it before commiting. Maybe this is just me but I often glance over it to see what was changed, especially when only bumping dependencies... 🤔

mnkhouri:
@LinusU doing a `git diff --text` should override the binary file filter. That doesn't help with `git add -p`, but maybe it's enough for visual inspection of package-lock.json? I don't feel strongly about the binary file filter, so if you find it to annoy you frequently, I'm totally ok with reverting

PR #272, opened 2018-08-13
Add export detection (`export x from y`)

### LINKED ISSUE #262: False "unused dependencies" for `export ... from ...`
I'm using a lot lines like this one:

```js
export { Foo as Bar} from 'some-package'
```
If a package is used only this way, depcheck will issue an "unused dependency" message. If I change it to this other equivalent code, then it won't detect as unused.

```js
import { Foo as Bar } from 'some-package'
export { Bar }
```

mnkhouri:
This PR fixes #262 by adding a detector for Export Declarations.


LinusU:
Nice 👍

mnkhouri:
tests are currently failing due to a `503 Service Temporarily Unavailable` from coveralls.io, I'll retry the build later

Issue #273, opened 2018-08-13
Dependencies out of date


mnkhouri:
Several of the project's dependencies are out of date, and `npm audit` contains several warnings. We should update dependencies before the next (major?) release

mnkhouri:
Resolved by #276

PR #274, opened 2018-08-13
Bump Mocha to 5.x

LinusU:
The only breaking changes are dropped support for runtimes that we aren't supporting so should be good to go. Fixes the most critical vulnerabilities reported by `npm audit`.

Related to #273


mnkhouri:
Excellent!

PR #276, opened 2018-08-17
Upgrade all dependencies

nkbt:
~It is WIP, since I wanted to check that all tests pass in CI (everything passes and works locally on my machine, I checked every script that CI runs) without updating code to match new eslint/airbnb~
All code is updated too, linting enabled back

- [x] Switch all old `babel-*` deps to new `@babel/*`
- [x] Replace deprecated packages with new ones
  - `babel-preset-*` ->`@babel/preset-env`
  - `codecov.io` -> `codecov`
  - `babel-node` -> `node -r @babel/register`
- [x] Upgrade all other dependencies to the most recent versions
- [x] Deprecate Node4 (many deps are incompatible now anyway) and add Node10
- [x] Update code to pass eslint + airbnb and uncomment lint check in CI. Done in a separate commit, so all changes can be easily diffed


src/cli.js
```diff
@@ -66,9 +66,9 @@ function print(result, log, json) {
 function checkDeprecation(argv) {
   if (argv.dev === false) {
     deprecate(
-      'The option `dev` is deprecated. It leads a wrong result for missing dependencies' +
-      ' when it is `false`. This option will be removed and enforced to `true` in next' +
-      ' major version.',
+      'The option `dev` is deprecated. It leads a wrong result for missing dependencies'
+      + ' when it is `false`. This option will be removed and enforced to `true` in next'
```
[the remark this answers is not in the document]
nkbt: Probably, I did not follow all their changes, just applied `--fix`. I used to be more involved with the code style stuff at some point, but found that I care less and less over time, as soon as it is automatic and"uniform" across the codebase

test/cli.js
```diff
@@ -94,16 +92,14 @@ describe('depcheck command line', () => {
...
+  it('should output error when folder not exists', () => testCli(['./not/exist/folder'])
     .then(({ log, error, exitCode }) => {
       error.should.containEql('/not/exist/folder').and.containEql('not exist');
       log.should.be.empty();
       exitCode.should.equal(-1);
     }));
 
-  it('should output call stack for invalid files in JSON view', () =>
-    testCli(makeArgv('bad_js', { json: true }))
+  it('should output call stack for invalid files in JSON view', () => testCli(makeArgv('bad_js', { json: true }))
```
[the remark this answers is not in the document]
nkbt: I guess it is out of scope for this PR, though I'm surely all for async/await these days.

mnkhouri:
Thanks for taking this on @nkbt! Fixes #273

nkbt:
- [x] Fixed eslint autofix for `} if () {`
- [x] Rebased

test/special/babel.js
```diff
@@ -98,35 +98,27 @@ describe('babel special parser', () => {
...
-      testBabel('package.json', testCase.deps, {
-        name: 'my-package',
-        version: '1.0.0',
-        babel: {
-          env: {
-            development: testCase.options,
-          },
-        },
-      })));
+  testCases.forEach(testCase => it(`should ${testCase.name} inside package.json file env section`, () => testBabel('package.json', testCase.deps, {
```
mnkhouri: The airbnb style calls for implicit returns [on the same line](https://github.com/airbnb/javascript#whitespace--implicit-arrow-linebreak) (as seen here). 

However, based on how the tests are written in this repo (with lots of nested implicit returns), that option makes them unreadable, so I think we should change the setting to ["implicit return below"](https://eslint.org/docs/rules/implicit-arrow-linebreak)

P.S. Sorry that this PR is turning into a style / eslint config discussion @nkbt  😞  I know it's a chore (but I agree that it's better to have the linter running / enforced!)

nkbt: No problem, I'll fix that

I would actually prefer to have `prettier` in place and just do not even think about style anymore. Basically as a common style that everyone equally dislikes but since it is automatic and you don't need to care about it - agree to follow =)

nkbt: I turned off that rule since there is a lot of code that follows either way

nkbt:
Reverted, turned off rule, re-applied autofix, manually applied some extra fixes for what linter could not do.

mnkhouri:
Thank you for getting this done, feels good to see the deps updated 😅 

I'll get this merged in the next few days, right after we get in the two bugfixes needed to release 0.6.10, #279 and #282 (thanks to your issue report on that one :)

nkbt:
Perfect, thank you!

mnkhouri:
@nkbt could you please also update the `README.md` line ["depcheck needs node >= 4"](https://github.com/depcheck/depcheck/blob/master/README.md#installation)?

nkbt:
Sure thing. Will try to get to my laptop this weekend =)

nkbt:
Updated.

mnkhouri:
Thank you @nkbt ! 🎉

nkbt:
Great, thank you!

PR #277, opened 2018-08-19
tests: remove Node.js 4, add Node.js 10

DanielRuf:


DanielRuf:
Weird, why does it fail here but not on previous commits? Is there some static check?

mnkhouri:
Weird, I'm not sure why that failure occurred! It looks like the same change will also be covered in #276 ([here, specifically](https://github.com/depcheck/depcheck/pull/276/files#diff-354f30a63fb0907d4ad57269548329e3L6)) -- so as long as the tests pass on that PR, I'm ok with not digging too deep into this failure

Anyway, thanks for submitting this PR :) PRs are always appreciated!

DanielRuf:
Right.

Saw it then after checking the other PRs.

Hopefully this will not occur with future PRs.

PR #279, opened 2018-08-20
Update NPM tokens with correct permissions

### LINKED ISSUE #275: Deploy failed (invalid NPM token?)
The deploy for version 0.6.10-beta.0 failed, see log [here](https://travis-ci.org/depcheck/depcheck/jobs/415614637#L2784). NPM [revoked all tokens 1 months ago](https://status.npmjs.org/incidents/dn7c1fgrr7ng), so I think our token needs to be updated

-----------------

For new token: https://www.npmjs.com/settings/lijunle/tokens

To [update travis](https://docs.travis-ci.com/user/encryption-keys/#usage): `travis encrypt api_key ="TOKEN VALUE" --add`

mnkhouri:
NPM package deployments are still failing, even with an updated token (see #275). One [issue comment](https://github.com/travis-ci/travis-ci/issues/9403#issuecomment-408641000) suggests that it is not sufficient to turn off 2fa-for-publishing after the token (as I did), but rather that the token must be created while 2fa-for-publishing is turned off.

Fixes #275

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ deploy:
     skip_cleanup: true
     email: npm@khouri.ca
     api_key:
-      secure: SlI8U43IpEctxgzP1H+GhxdxPE2VNNuSfA85YZdnlQ/gMQ2hifZQYP1q9k3z2H4RRdo+ooEkmcsXE0a/s6XGrTIO98p4zI2LapOsEgedNsYwBWmLSwq0r2LCXDc5btvr043NdPFkhhfhNcmCtYFuWrAMLjg4ypttBcbMdMPj+Jw=
+      secure: EuVF1Ji3RIO6JF91ww4GIx5qUY2xiaPUfYA92uvqiGwJam7taL/v3OLy6Y5khjnXxFKgfkjd+x2tQ9n2hS56t9Ky7ud8Zw439an/PqEWGE3olTv4n7ZdVaAHzcIydlNL5qmezdlTDM+tmggD7gWWsjatYc9QY4fSnGUtdRZ4HzM=
     on:
       tags: true
       node: 6


lijunle:
Have you tried to manually publish a new version locally?

mnkhouri:
Using this same token locally in my .npmrc works 👍 (just [published `0.6.10-beta.1` manually](https://www.npmjs.com/package/depcheck/v/0.6.10-beta.1) to test)

.travis.yml
```diff
@@ -25,7 +25,7 @@ deploy:
     skip_cleanup: true
     email: npm@khouri.ca
     api_key:
-      secure: SlI8U43IpEctxgzP1H+GhxdxPE2VNNuSfA85YZdnlQ/gMQ2hifZQYP1q9k3z2H4RRdo+ooEkmcsXE0a/s6XGrTIO98p4zI2LapOsEgedNsYwBWmLSwq0r2LCXDc5btvr043NdPFkhhfhNcmCtYFuWrAMLjg4ypttBcbMdMPj+Jw=
+      secure: EuVF1Ji3RIO6JF91ww4GIx5qUY2xiaPUfYA92uvqiGwJam7taL/v3OLy6Y5khjnXxFKgfkjd+x2tQ9n2hS56t9Ky7ud8Zw439an/PqEWGE3olTv4n7ZdVaAHzcIydlNL5qmezdlTDM+tmggD7gWWsjatYc9QY4fSnGUtdRZ4HzM=
```
nkbt: You need to use UI to add secured tokens, do not expose tokens in source code

https://blog.travis-ci.com/2018-04-11-how_to_publish_node_js_packages_with_travis_ci_and_packagecloud/

mnkhouri: @nkbt this line is using [Travis's encryption](https://docs.travis-ci.com/user/encryption-keys/) to hide the API keys. Based on the Travis documentation, only the CI server will be able to decrypt this key 🔑 so I think we're in good shape here

nkbt: I've seen that in docs. But does it mean that this key is encrypted per account? So if you are using different account to encrypt it - it will not work?

Either way for better security I would avoid keeping any tokens in github. Encrypted or not. Especially if there is an easier way with encrypted ENV vars.

mnkhouri: Based on the docs, the encryption key is per-repo. Since this method is documented in Travis's docs, I don't see a security issue with it. For UI vs travis.yml, I'd prefer to keep the key in travis.yml, since then there is only one single point of config for Travis (rather than travis.yml + some changes in the UI)

PR #283, opened 2018-08-26
Add tests for esnext syntax in Typescript

mnkhouri:
These tests cover the changes in PR #258 (they fail without PR #258).


mnkhouri:
I should have asked for these tests as part of PR #258, but better late than never :)

PR #284, opened 2018-09-02
Improve regex in `parseLinter()`

hassankhan:
The regex now allows for files like `tslint.json` as well as `.tslintrc`.


mnkhouri:
Hi @hassankhan, in case you want to try it out, this PR is now published on npm in the `depcheck@next` tag (in version `0.7.0-alpha.1`). This is a pre-release version, `0.7.0` will be officially published in a few weeks.

Thanks again for the PR

PR #286, opened 2018-09-03
Upgrade NPM in CI to a version that uses package-lock.json

mnkhouri:
CI is currently using npm@3 in AppVeyor, and npm@3 in Travis's Node v6 env
(because npm@3 is bundled with Node v6). npm@3 does not use package-lock.json.

This has caused an issue in CI, because ["deprecate@**1.1.0**" was installed](https://ci.appveyor.com/project/lijunle/depcheck/build/519/job/5c7qv724u4hqtoh4#L447)
and broke builds [due to this issue](https://github.com/brianc/node-deprecate/pull/7), even though "deprecate@**1.0.0**" is
specified in our package-lock.json.

npm@5 is the first version that uses package-lock.json by default, but I think
there's no harm in just using npm@6.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,9 +10,6 @@ node_js:
 before_install:
   - dpkg --compare-versions `npm -v` ge 6 || npm i -g npm@^6
 
-install:
-  - npm ci
-
 script:
   - npm run compile
   - npm run component


appveyor.yml
```diff
@@ -6,7 +6,7 @@ environment:
 
 install:
   - ps: Install-Product node $env:nodejs_version
-  - npm -g install npm@3 && set PATH=%APPDATA%\npm;%PATH%
+  - npm -g install npm@6 && set PATH=%APPDATA%\npm;%PATH%
   - npm install
```
LinusU: Can we change this to `npm ci`? That way it will fail the build if the `package-lock.json` file isn't up to date with the `package.json` file

mnkhouri: Good idea 👍 just committed this change for both travis and appveyor

.travis.yml
```diff
@@ -7,7 +7,11 @@ node_js:
   - 8
   - 10
 
-before_install: dpkg --compare-versions `npm -v` ge 6 || npm i -g npm@^6
+before_install:
+  - dpkg --compare-versions `npm -v` ge 6 || npm i -g npm@^6
+
+install:
+  - npm ci
```
LinusU: Travis already does this automatically ☺️ 

https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#npm-ci-support

.travis.yml
```diff
@@ -7,6 +7,9 @@ node_js:
   - 8
   - 10
 
+before_install:
+  - dpkg --compare-versions `npm -v` ge 6 || npm i -g npm@^6
```
LinusU: Just realized that it _might_ be a good idea to use the older npm when testing Node.js 6, since that is what people using Node.js 6 will be using. That way we would catch errors/incompatibilities with older npm versions.

We would still get the benefit of getting the `package-lock.json` checked since `npm ci` would be used in the newer Node.js versions.

Not a big deal though, I'm fine either way. Could be nice with a bit less code though ☺️

LinusU: Hmm, just read the initial PR description as well, this is probably good then 👍

Issue #188, opened 2016-12-01
Unused devDependencies with Karma


screendriver:
I am using Karma and Webpack for my unit tests. Unfortunately depcheck `v0.6.6` tells me

```sh
Unused devDependencies
* babel-core
* babel-loader
* babel-polyfill
* karma-chrome-launcher
* karma-coverage
* karma-mocha
* karma-mocha-reporter
* karma-teamcity-reporter
* karma-webpack
* mocha
* webpack
```
 
### karma.conf.js

```js
// Karma configuration
// Generated on Thu Aug 18 2016 11:47:04 GMT+0200 (CEST)

module.exports = (config) => {
  config.set({

    // base path that will be used to resolve all patterns (eg. files, exclude)
    basePath: '',


    // frameworks to use
    // available frameworks: https://npmjs.org/browse/keyword/karma-adapter
    frameworks: ['mocha'],


    client: {
      mocha: {
        // change Karma's debug.html to the mocha web reporter
        reporter: 'html',
      },
    },


    // list of files / patterns to load in the browser
    files: [
      'node_modules/babel-polyfill/dist/polyfill.js',
      { pattern: 'test/**/*.test.js' },
    ],


    // list of files to exclude
    exclude: [
    ],


    // preprocess matching files before serving them to the browser
    // available preprocessors: https://npmjs.org/browse/keyword/karma-preprocessor
    preprocessors: {
      'test/**/*.test.js': ['webpack'],
    },


    webpack: {
      devtool: 'inline-source-map',
      externals: [{
        chcp: true,
        device: true,
      }],
      module: {
        loaders: [{
          test: /\.js$/,
          exclude: /node_modules/,
          loader: 'babel',
        }],
      },
    },


    // test results reporter to use
    // possible values: 'dots', 'progress'
    // available reporters: https://npmjs.org/browse/keyword/karma-reporter
    reporters: (process.env.TEAMCITY_VERSION
       ? ['teamcity']
       : ['progress', 'mocha'])
       .concat(['coverage']),


    coverageReporter: {
      reporters: [
         { type: 'html', dir: 'coverage/', subdir: '.' },
         { type: 'text-summary' },
      ].concat(process.env.TEAMCITY_VERSION ? [{ type: 'teamcity' }] : []),
    },


    // web server port
    port: 9876,


    // enable / disable colors in the output (reporters and logs)
    colors: true,


    // level of logging
    // possible values: config.LOG_DISABLE || config.LOG_ERROR ||
[...]

PR #306, opened 2019-01-23
Karma support

rjatkins:
Implements #188

### DIFF
--- a/src/special/karma.js
+++ b/src/special/karma.js
@@ -155,7 +155,7 @@ function collectReporters(config, pluginInfo) {
   const reporterMapping = collectInstalledPluginOfType('reporter', pluginInfo);
   const installedReporters = Object.keys(reporterMapping);
   // config defines a property reporters: ['<name>', ...]
-  return Object.values(wrapToMap(config.reporters))
+  return wrapToArray(config.reporters)
     .filter(name => installedReporters.includes(name))
     .map(name => reporterMapping[name]);
 }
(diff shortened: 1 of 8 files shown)


src/special/karma.js
```diff
@@ -0,0 +1,160 @@
...
+  return obj;
+}
+
+function collectFrameworks(config, karmaPluginsInstalled) {
+  // karma-x plugins define frameworks using the structure { 'framework:<name>': ['factory', f] }
+  // generate a lookup map { '<name>': 'karma-x', ... }
+  const frameworkMapping = {};
+  // skip plugins with more specific names, such as launchers and reporters
+  karmaPluginsInstalled.filter(plugin => !plugin.endsWith('-launcher') && !plugin.endsWith('-reporter')).forEach((genericPlugin) => {
+    const p = tryRequire(genericPlugin);
```
rjatkins: This specific approach - using node's require to try to pull in the relevant module - seems to be against the philosophy of the rest of the project. Sadly the information we need to resolve frameworks/browsers/reporters/etc to plugins isn't quite available statically, which means we need to find the default module in each plugin and parse out the map from its module.export. Any guidance on how to do that cheaply would be much appreciated!

rumpl:
The build fails on node 6, I'm not against dropping node 6 support...

src/special/karma.js
rjatkins: Opted to replace it with parser code, so I can guarantee it works in my own project.

rjatkins:
> The build fails on node 6, I'm not against dropping node 6 support...

It's easy enough to stop using `Object.values` and make it backwards compatible again, though I'm really confused why `@babel/polyfill` isn't triggering in appveyor, but works fine in travis. Doing that instead.

rumpl:
🎉

Thanks!

PR #307, opened 2019-01-25
Replace babylon with @babel/parser, and upgrade all babel dependencies

rjatkins:
* Note that @babel/parser 7+ no longer supports wildcard plugins config, which forces us to be explicit. See https://babeljs.io/docs/en/babel-parser#plugins for the documented list of plugins


lijunle:
@rjatkins thank you very much on contributing on this project. Any breaking changes between babylon and @babel/parser?

rumpl:
I'm pretty much OK with flow being disabled :)

Thank you so much for contributing!

PR #309, opened 2019-01-28
correct resolution of lint `extends` packages

43081j:
There's currently a bug in master in that extending `plugin:@scope/name` should result in `@scope/eslint-plugin` but we incorrectly resolve it as `@scope/eslint-plugin-name` or some such thing.

In this PR, I have removed most of what logic we had around `extends` and (like ESLint does its self) have used the normalisation function I recently introduced.

I also removed a bunch TSLint tests, because TSLint only supports loading inherited configs by node resolution (it just does a `require` on it). Because of this, I had to add a short-circuit in the resolution function to return the path as-is for tslint.

cc @rumpl

### DIFF
--- a/src/special/tslint.js
+++ b/src/special/tslint.js
@@ -18,6 +18,13 @@ function checkConfig(config, rootDir) {
     .map(requirePackageName);
 }
 
+/**
+ * Parses TSLint configuration for dependencies.
+ *
+ * TSLint uses node resolution to load inherited configurations.
+ * More info on this can be found
+ * [here](https://palantir.github.io/tslint/usage/configuration/).
+ */
 export default function parseTSLint(content, filename, deps, rootDir) {
   const config = loadConfig('tslint', filename, content);
   if (config) {
--- a/src/special/tslint.js
+++ b/src/special/tslint.js
@@ -1,5 +1,28 @@
-import parseLinter from '../utils/linters';
+import path from 'path';
+import requirePackageName from 'require-package-name';
+import { loadConfig } from '../utils/linters';
+import { wrapToArray } from '../utils/index';
+
+function resolvePresetPackage(preset, rootDir) {
+  if (preset.startsWith('./') || preset.startsWith('../')) {
+    return path.resolve(rootDir, preset);
+  }
+  return preset;
+}
+
+function checkConfig(config, rootDir) {
+  return wrapToArray(config.extends)
+    .filter(preset => !preset.startsWith('tslint:'))
+    .map(preset => resolvePresetPackage(preset, rootDir))
+    .filter(preset => !path.isAbsolute(preset))
+    .map(requirePackageName);
+}
 
 export default function parseTSLint(content, filename, deps, rootDir) {
-  return parseLinter('tslint', content, filename, deps, rootDir);
+  const config = loadConfig('tslint', filename, content);
+  if (config) {
+    return checkConfig(config, rootDir);
+  }
+
+  return [];
 }
--- a/src/utils/linters.js
+++ b/src/utils/linters.js
@@ -1,10 +1,8 @@
 import path from 'path';
 import yaml from 'js-yaml';
-import lodash from 'lodash';
-import requirePackageName from 'require-package-name';
-import { evaluate, wrapToArray } from '.';
+import { evaluate } from '.';
 
-function parse(content) {
+export function parse(content) {
   try {
     return JSON.parse(content);
   } catch (error) {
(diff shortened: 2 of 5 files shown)


src/utils/linters.js
```diff
@@ -48,6 +48,10 @@ function isLinterConfigFromAFullyQualifiedModuleName(specifier, prefix) {
 }
 
 function resolvePresetPackage(flavour, preset, rootDir) {
+  if (flavour === 'tslint') {
```
43081j: i needed to do this because tslint only supports node resolution. if there's a better place or way to do this, do let me know

rumpl:
Hey, thanks for this, I'll take a look at it tomorrow, it's late here now.

Thanks for contributing 🎉🎉🕺🎉🎉

src/utils/linters.js
```diff
@@ -48,6 +48,10 @@ function isLinterConfigFromAFullyQualifiedModuleName(specifier, prefix) {
...
+    return preset;
+  }
+
   // inspired from https://github.com/eslint/eslint/blob/5b4a94e26d0ef247fe222dacab5749805d9780dd/lib/config/config-file.js#L347
```
lijunle: Do we have a new reference?

lijunle: Worth to mention the TSLint resolution code as a comment.

43081j: Sure, just the main docs specify it, it hasn't changed we just always implemented it incorrectly.

I'm still not sure this is the best thing to do though, it feels a bit hacky having a tslint-specific short circuit here. will have another think in case i have a better idea

lijunle: Oh, I point to the line about `// inspired from https://`, not your fast return.

BTW, I suppose TSLint still support preset with syntax `tslint:latest`.

43081j: Yes I know, I just meant in addition to that any suggested alternatives would be nice :)

As for what they support, yes, `tslint:recommended` and `latest` are too. But we already do this wrong, we only check for `recommended` in both es and tslint (we should just check `{flavour}:*` i guess

lijunle: See the example here: https://palantir.github.io/tslint/usage/configuration/#configuration-presets

lijunle: I see. I think divorce the logic between ESLint and TSLint is the correct way.

43081j:
alright here we go.

tslint config is really really simple, it turns out we were implementing a whole bunch of stuff tslint doesn't even support. the only thing it supports is `extends` with a module or path really. 

so i went ahead and split it out into its own implementation.

the tests have been cleaned up to be accurate to what tslint actually has too.

rumpl:
Dude, <3

Thank you :)

rumpl:
Would you like to be part of the “core team”?

43081j:
Sure, if you'd like. Would be happy to have a stab at some other issues when i'm free if needed.

src/special/tslint.js
```diff
@@ -1,5 +1,28 @@
...
+
+function checkConfig(config, rootDir) {
+  return wrapToArray(config.extends)
+    .filter(preset => !preset.startsWith('tslint:'))
+    .map(preset => resolvePresetPackage(preset, rootDir))
+    .filter(preset => !path.isAbsolute(preset))
+    .map(requirePackageName);
+}
 
 export default function parseTSLint(content, filename, deps, rootDir) {
```
lijunle: Could you please paste this link somewhere around here?

> https://github.com/palantir/tslint/blob/95d9d958833fd9dc0002d18cbe34db20d0fbf437/src/configuration.ts#L289

test/special/tslint.js
```diff
@@ -11,36 +11,20 @@ const testCases = [
     expected: [],
   },
   {
-    name: 'detect specific plugins',
+    name: 'skip single built-in config',
```
lijunle: I suppose the second one have covered this. We can remove this one.

test/special/tslint.js
```diff
@@ -11,36 +11,20 @@ const testCases = [
...
+    name: 'skip built-in configs',
     content: {
-      extends: 'tslint:recommended',
+      extends: [
+        'tslint:recommended',
+        'tslint:latest',
+        'tslint:foo',
+      ],
     },
     expected: [],
```
lijunle: I am thinking about if we should declare `tslint` in `expected` array. I am not sure if there is a user case to declare the `tslint.json` but not `script` CLI command. I personally think it is worth to declare `tslint` here.

lijunle:
@43081j The changes look good! Thank you very much for the contribution! BTW, I have invited you to the team. Welcome!

test/special/tslint.js
43081j: this is to test that you can use `extends: 'foo'` rather than an array, `extends: ['foo']`. so i think we should leave it here?

test/special/tslint.js
43081j: just to clarify, does this mean the tslint parser would return `['tslint']` every time (at least)?

src/special/tslint.js
43081j: i added jsdoc linking to the docs instead as it explains there in friendlier terms what resolution is used.

PR #311, opened 2019-02-03
cleanup mocha implementation

### LINKED ISSUE #241: False positive when using vuegister with mocha
My package.json:

```json
{
  "scripts": {
    "test": "mocha --require babel-register --require vuegister/register tests/*.spec.js"
  },
  "dependencies": {
    "vue": "^2.5.13"
  },
  "devDependencies": {
    "babel-core": "^6.26.0",
    "babel-loader": "^7.1.2",
    "depcheck": "^0.6.8",
    "mocha": "^5.0.0",
    "vuegister": "^0.2.2",
    "vuegister-plugin-babel": "^0.2.1"
  }
}
```

`depcheck --json` output:

```json
{
  "devDependencies": [
    "vuegister",
    "vuegister-plugin-babel"
  ]
}
```

### LINKED ISSUE #287: false alert on mocha-jenkins-reporter (cant seem to add `help wanted` label)
mocha-jenkins-reporter is reported as being unused.

npx depcheck --json:
```
{
  "dependencies": [],
  "devDependencies": [
    "mocha-jenkins-reporter"
  ],
  "missing": {},
  "using": {
    "depcheck": [
      "/Users/carrie.prebble/Downloads/depcheck-mocha-jenkins-reporter/package.json"
    ],
    "mocha": [
      "/Users/carrie.prebble/Downloads/depcheck-mocha-jenkins-reporter/package.json"
    ],
    "chai": [
      "/Users/carrie.prebble/Downloads/depcheck-mocha-jenkins-reporter/someCode.spec.js"
    ]
  },
  "invalidFiles": {},
  "invalidDirs": {}
}
```

When I remove this dependency and run my tests, I get the following error:
```
"mocha-jenkins-reporter" reporter not found
/Users/carrie.prebble/source/share-my-data/node_modules/mocha/lib/mocha.js:193
      throw new Error('invalid reporter "' + reporter + '"');
      ^
Error: invalid reporter "mocha-jenkins-reporter"
    at Mocha.reporter (/Users/carrie.prebble/source/share-my-data/node_modules/mocha/lib/mocha.js:193:13)
...
```

mocha-jenkins-reporter is used for Node.js backend tests using mocha to output nicely formatted Jenkins reports of the test runs

It's used in our package.json test script:
```
"test": "JUNIT_REPORT_PATH=./jenkins-test-results.xml mocha \"*.spec.js\" --colors --reporter mocha-jenkins-reporter"
```

The attached project is a trivial example to reproduce the false alert issue.

To reproduce:
1. unpack depcheck-mocha-jenkins-reporter.zip
2. yarn
3. yarn package

[depcheck-mocha-jenkins-reporter.zip](https://github.com/depcheck/depcheck/files/2353630/depcheck-mocha-jenkins-reporter.zip)

43081j:
* Fixes #287
* Fixes #241 

So this should help out with fixing up the mocha implementation a bit. Did some refactoring, added support for `--reporter` and `--require` in both opts files and scripts.

Wasn't too sure if the tests were good enough so some feedback on them would be appreciated.

If there's anything obvious i've messed up do let me know.


src/utils/get-scripts.js
```diff
@@ -5,6 +5,12 @@ import lodash from 'lodash';
 
 const scriptCache = {};
 
+export function clearCache() {
+  Object.keys(scriptCache).forEach((key) => {
```
lijunle: Why not declare `scriptCache` as `let` and directly `= {}` as `clearCache`?

src/special/mocha.js
```diff
@@ -1,41 +1,82 @@
...
+    return null;
+  }
+
+  const optsPath = argvs[optsIndex + 1];
+
+  if (!optsPath) {
+    return null;
+  }
+
+  return fs.readFileSync(path.resolve(root, '..', optsPath), 'utf-8');
```
lijunle: What if the file does not exist? The exception will break whole parser.

src/special/mocha.js
```diff
@@ -1,41 +1,82 @@
...
+
+  return result
     .map(requirePackageName)
-    .filter(name => deps.indexOf(name) !== -1);
+    .filter((v, k, arr) => arr.indexOf(v) === k)
+    .filter(name => deps.includes(name));
 }
 
 export default function parseMocha(content, filepath, deps, rootDir) {
   const defaultOptPath = path.resolve(rootDir, 'test/mocha.opts');
```
lijunle: I don't really recall how mocha config works. It could better to get one more review if you are not fully confident. The test cases look good to me.

src/special/mocha.js
```diff
@@ -1,41 +1,82 @@
 import fs from 'fs';
 import path from 'path';
-import lodash from 'lodash';
 import requirePackageName from 'require-package-name';
 import { getScripts } from '../utils';
 
-function getOpts(script) {
-  const argvs = script.split(' ').filter(argv => argv);
+const knownReporters = [
+  'dot', 'doc', 'tap', 'json', 'html', 'list',
```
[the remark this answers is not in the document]
rumpl: +1 on prettier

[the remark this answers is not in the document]
lijunle: I am fine to check in this PR first then enable prettier.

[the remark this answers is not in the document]
43081j: yup this isn't actually how i would style my code either. i had left it this way as i was meant to suggest we use prettier and may have forgot... 🙈

src/utils/get-scripts.js
43081j: so we can maintain the same reference/object. it doesn't make a difference right now, but if for example we added a test to check the cache explicitly, we'd want to keep the same reference outside this file.

src/special/mocha.js
43081j: what would be the correct thing to do? if they pass `--opts nonexistent`, we should throw or error in some way so they know they've tried to pass a non-existent config.

src/special/mocha.js
```diff
@@ -1,41 +1,82 @@
...
+        result.push(val);
+      }
+    }
+  });
+
+  return result
     .map(requirePackageName)
-    .filter(name => deps.indexOf(name) !== -1);
+    .filter((v, k, arr) => arr.indexOf(v) === k)
+    .filter(name => deps.includes(name));
```
43081j: this seems weird, maybe it shouldn't be here? it was there before but i wasn't sure why...

rumpl:
I’ll take a look tomorrow, it’s late where I am

src/special/mocha.js
43081j: it reads as if it will filter out any dependencies which aren't in our package.json? don't we want to know about those? i left it in because it was there before but seems pretty questionable.

Issue #319, opened 2019-02-12
Not working for Typescript Angular 7 project


zentoaku:
Hello,

I'm trying to figure out how to use this tool with Angular 7 project which uses Typescript. I've seen that there should be a built-in support for it but so far the tool is reporting all of my dependencies from package.json as "unused". 
Tried:
> depcheck ./
> depcheck ./ --parsers="*.ts:ts"
> depcheck ./ --parsers="*.ts:typescript"

Could you give me some more clarity on how to use this tool from command line?

rumpl:
Hi @zentoaku,

I just tried this:
* install ng-cli: `npm install -g @angular/cli`
* create a new app: `ng new my-app`
* run depcheck: `depcheck .`

And this is the output I get: 
```
Unused dependencies
* @angular/animations
* @angular/forms
* core-js
* tslib
Unused devDependencies
* @angular/cli
* @angular/compiler-cli
* @angular/language-service
* @types/node
* @types/jasmine
* @types/jasminewd2
* codelyzer
* tslint
* typescript
```

The unused dependencies are correct as for the devDependencies, there is a bunch of stuff going on:
* the `@types/*` are not yet supported by depcheck, should be there soon
* not sure about the cli or compiler-cli
* tslint, typescript and codelyzer seem like a bug

Could you perhaps share a repo against which you are running depcheck? Or at least give us the output of `depcheck . --json`.

Thanks

zentoaku:
I can't share the repo but here is the command output:
```
{
  "dependencies": [
    "@angular/animations",
    "@angular/common",
    "@angular/compiler",
    "@angular/core",
    "@angular/forms",
    "@angular/http",
    "@angular/platform-browser",
    "@angular/platform-browser-dynamic",
    "@angular/router",
    "@xxx/breadcrumbs",
    "@xxx/cache",
    "@xxx/globals",
    "@xxx/notification",
    "@xxx/security",
    "@xxx/translation",
    "@xxx/web-storage",
    "@ng-bootstrap/ng-bootstrap",
    "@ngrx/effects",
    "@ngrx/router-store",
    "@ngrx/store",
    "@ngrx/store-devtools",
    "@ngx-translate/core",
    "@progress/kendo-angular-buttons",
    "@progress/kendo-angular-dateinputs",
    "@progress/kendo-angular-dropdowns",
    "@progress/kendo-angular-excel-export",
    "@progress/kendo-angular-grid",
    "@progress/kendo-angular-inputs",
[...]

PR #325, opened 2019-03-04
bugfix: include tslint in deps when a tslintrc exists

43081j:
This should fix part of a few issues like #319.

I introduced `tslint` as a dependency every time a `tslintrc` file exists. Let me know if you think this is reasonable (and also if its fine that we still expect it when the tslint config is invalid).

### DIFF
--- a/src/special/eslint.js
+++ b/src/special/eslint.js
@@ -102,8 +102,10 @@ function checkConfig(config, rootDir) {
   return lodash.union(parser, plugins, presetPackages, presetDeps);
 }
 
+const configNameRegex = /^\.eslintrc(\.(json|js|yml|yaml))?$/;
+
 export default function parseESLint(content, filename, deps, rootDir) {
-  const config = loadConfig('eslint', filename, content);
+  const config = loadConfig(configNameRegex, filename, content);
   if (config) {
     return checkConfig(config, rootDir);
   }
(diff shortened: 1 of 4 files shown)


43081j:
not too sure whats up here.

the `build/check-json` task ends up thinking our `src/special/tslint.js` requires tslint by the looks of it. maybe i went wrong somehow

43081j:
alright so i've made a change to the config file regex matching, too.

it turns out our patterns were _kinda_ wrong..

ESLint supports: `.eslintrc*` (yaml, json, yml)
TSLint supports `tslint.(json|yaml|yml)` (**there is no concept of `.tslintrc` etc**)

should this be considered a breaking change? as we will no longer support those unsupported filenames.

rumpl:
Huh, weird that we thought that there was a `.tslintrc`...
We should keep in mind that https://github.com/palantir/tslint/issues/4534

rumpl:
Could you please rebase so that we can merge safely, thanks!

43081j:
I suspect it may have been supported a very long time ago, or could just be an easy typo as some aged projects do have such a file. 

Tslint deprecation is great news. Eslint already supported typescript for a long time now. 

Will rebase soon too

43081j:
@rumpl rebased

rumpl:
Merged 🎉

PR #333, opened 2019-03-20
support babel.config.js

dword-design:
In Babel 7 it is common practice to have a project-wide babel.config.js, so we should support it


rumpl:
Thanks for the PR, could you fix the linter erros so that we can merge?

dword-design:
Sorry missed that. Here you go

jrnail23:
In order to properly handle Babel 7, this PR also needs to address babel's plugin and preset renaming (i.e., `babel-plugin-something` is now `@babel/plugin-something`).
see: https://babeljs.io/docs/en/v7-migration#scoped-packages

dword-design:
@jrnail23 hm I don't know exactly what you mean. The scoped packages should be put into babel.config.js, but this shouldn't be related to parsing.

rumpl:
Thanks @dword-design

PR #335, opened 2019-03-20
add babel plugins to vue special

dword-design:
Since they were missing so far :)

One thing to note about this:
It would probably be more flexible to use @babel/core.parseSync() instead of @babel/parser.parse(), because @babel/core implicitly looks for a babel.config.js and uses that for parsing. This way we don't have to brute force add all plugins to the config.


rumpl:
Thanks!

PR #336, opened 2019-03-21
Allow process var access in linter configs

dword-design:
The current evaluation function does not support Node.js features like process.env.NODE_ENV, which is helpful when e.g. defining configs that depend on the environment. This PR adds require-from-string, which supports these features.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,6 +57,7 @@
     "minimatch": "^3.0.2",
     "node-sass-tilde-importer": "^1.0.2",
     "please-upgrade-node": "^3.1.1",
+    "require-from-string": "^2.0.2",
     "require-package-name": "^2.0.1",
     "resolve": "^1.10.0",
     "walkdir": "0.0.12",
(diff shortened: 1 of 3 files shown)


GarrettGeorge:
@dword-design Personally (I'm not an active contributor to this project), I would steer away from ditching `evaluate`. That function uses node's native `vm` module to sandbox the execution of the code.

I think a better approach is to have a `depcheck` config file which would specify certain parameters for the `--specials`. This way you could define `globals` similar to how [jest does it](https://jestjs.io/docs/en/configuration#globals-object).

An alternative, if this is affecting your own codebase, is to create a `dev` and `prod` `.eslintrc`. Obviously this isn't the best approach, but it prevents premature PRs that `depcheck` would have to change in the future.

GarrettGeorge:
An alternative is to add temporary support **just** for a `--process-env` flag which is injected into the sandbox for `evaluate`.

It's possible a `--process-flag` ends up being useful for more than just this use case. May be worth it to make a PR with that flag added.

rumpl:
The evaluation of config files was a good idea but it just doesn’t work. Best thing to do is like you said: a depcheck config file where people will put their stuff.

PR #355, opened 2019-06-10
Extract base package names for webpack

dword-design:
Currently the webpack special takes the full loader string as a package name. In fact, one can also pass subpackages like 'linaria/loader' as loaders. This PR uses require-package-name to extract the base package name out of the strings.

### DIFF
--- a/src/special/webpack.js
+++ b/src/special/webpack.js
@@ -1,6 +1,6 @@
 import path from 'path';
 import lodash from 'lodash';
-import requirePackageName from 'require-package-name'
+import requirePackageName from 'require-package-name';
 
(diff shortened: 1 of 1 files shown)


rumpl:
Don't we [already do this?](https://github.com/depcheck/depcheck/blob/beae930a05b2e00f8123e28a635ebe4c25b0f8ec/src/check.js#L73)

rumpl:
If not, it would be great if you could add some tests for this.

dword-design:
@rumpl Indeed this seems to have been added recently and my fork was outdated. So this can be closed :).

PR #408, opened 2019-08-10
Support loading configuration from a config file.

Urik:
Added an option to load ignore rules from a lines separated file by using the "--ignores-file"
CLI flag.

Usage would be `depcheck --ignores-file="pathToFile"`

I wanted to add unit tests but could find no mocking libraries installed in the system to handle "rs.readFile" calls, nor was I sure of exactly where should I add my unit tests.
Any help in that front will be very welcome.


src/utils/read-lines-separated-file.js
```diff
@@ -0,0 +1,13 @@
+import { readFile } from 'fs';
+
+export default function readLinesSeaparatedFile(filePath) {
+  return new Promise((resolve, reject) => {
+    readFile(filePath, 'utf-8', (err, content) => {
+      if (err) {
+        return reject(err);
+      }
+
+      return resolve(content.split('\n'));
```
rumpl: Could we use a regular expression here, on windows a new line is `\r\n`.

rumpl:
Hey @Urik 

Thanks for the PR, I'll try to find some time today and see how we could add unit tests for this.

rumpl:
@Urik I just pushed a test for this [feature in another branch](https://github.com/depcheck/depcheck/commit/1efb6dead0e07ed7dcae37830c923202b4dc0aab), you can copy/paste the test in your branch and then we'll be ready to merge :)

src/utils/read-lines-separated-file.js
Urik: I was under the impression fs.readLine always interprets `\r\n` as simply `\n` but now I couldn't find anything about it. Fixed.

Urik:
Thank you for the tests!
Just added them, so the code should be ready for review again.

rumpl:
This looks good and I'm really close to merging it but it got me thinking, why stop here?

There were people asking for a .depcheckrc (or whatever) file so that it's more CI friendly.

What do you think @Urik ? Are you motivated enough to change this PR to a full-fledged depcheck config file? I don't think it would be too much work...

Urik:
Mmmh I agree it'd be a very nice addition and a .depcheckrc file would make life much easier for me for sure, but it would still be nice to provide the option to have the ignore rules in a separate file, away from other config options, making it easier to separate the concern of the ignore rules from the rest of the configuration.

I can work on it on the following weekends, but it'll take me some time as I'm very busy on the weekdays and it'll take a few hours. I think it'd be best handled as a separate feature and pull request.

rumpl:
@KaboomFox @lijunle @mnkhouri @43081j 

Any thoughts on this?

43081j:
in most projects ive seen, there wouldn't usually be a "load ignores from file" feature.

normally you'd have a `*rc.{json,yml}` file which can contain a `ignore: [a, b, c]` while the CLI equivalent is `--ignore=a,b,c`.

personally id prefer this approach rather than supporting loading the list from its own standalone file, for sake of keeping it well organised and to a 'convention' (in that everything is guaranteed to be on the command line or in a config file, nowhere else).

but if you're all up for having this feature, i'd do it along side the rc file implementation (json only in first cut).

mnkhouri:
I'm also in agreement: if we have some configuration available via a file (as this PR implements, which I think is a great idea!) we should have all configuration eventually available via file.

I've seen a lot of other projects use [cosmiconfig](https://github.com/davidtheclark/cosmiconfig), which implements the logic for finding and parsing config files. It seems pretty standard among other javascript cli tools (e.g. babel, eslint, prettier).

I've never personally developed with cosmiconfig, but the documentation is good and the package is well-tested, so perhaps it would be a reasonable lift to make use of cosmiconfig with this PR? We could start with just supporting `ignores` in the config file, then expand in other PRs

Urik:
My rationale was that ignores rules might change more often than not while config values wouldn't, thus it would be nice to be able to easily tell away changes between a proper config value and one of the possibly many ignores. 

I like @mnkhouri 's approach of adding support for other settings progressively.

lijunle:
I like the `*rc` approach in general. It will be very help if @Urik can make that work.

Besides, may I step back and ask what kind of packages do you want to ignore? You said that it is related to unit test, but with what concrete case?

Urik:
I finally had a free Sunday and implemented the RC file functionality. Thanks @mnkhouri for the cosmiconfig suggestion, it was of great help.

Something I haven't had time to implement yet is supporting comma separated lists in CLI arguments, and real JSON arrays in the .depcheckrc file, since that'd mean changing more parts of the logic and I had no time.

Is it something you want supported, or would you like to keep the comma separated notation for the sake of consistency and simplicity?

Urik:
Sorry it took me so long, I finally had time today to implement this. 
The feature is ready for review. It'll mimick the cli args, meaning you can use the same arguments as in the CLI arguments. 
#### Priority
If an argument is found both in the CLI and in the .depcheckrc file, the CLI argument will take precedence, overwriting the value in the .depcheckrc.
#### Format
the .depcheckrc file can use YAML notation:    
```
ignores: ["a", "b"]
ignore-dirs:
  - folderA
  - folderB
```
or also JSON notation:
```
{
  "ignores": ["a", "b"],
  "ignore-dirs": ["folderA", "folderB"]
}
```

Also to take into account, keys can be `camelCase`, or also `kebab-case`, it'll have no effect.

#### Extension
Thanks to cosmiconfig, the .depcheckrc file can have no extension, `.json` extension, `.yml` extension, `.yaml` extension, and `.js` extension.

rumpl:
> Sorry it took me so long, I finally had time today to implement this.

Please don't be sorry, this is awesome! :)

README.md
```diff
@@ -79,6 +79,19 @@ All of the arguments are optional:
 
 `--parsers`, `--detectors` and `--specials`: These arguments are for advanced usage. They provide an easy way to customize the file parser and dependency detection. Check [the pluggable design document](https://github.com/depcheck/depcheck/blob/master/doc/pluggable-design.md) for more information.
 
+## Usage with a configuration file
+
+Depcheck can be used along with rc configuration files. In order to do so, create a .depcheckrc file in your project's package.json folder, and set the CLI keys in YAML format.
+For example, the CLI arguments `--ignores="eslint,babel-*" --skip-missing=true` would turn into:
```
43081j: worth mentioning here what formats are supported rather than just YAML (as i think you said it can do JSON too?)

Urik: Good catch, updated

azai91:
Any idea when this will be merged?

rumpl:
Tomorrow, thanks for pinging me, I forgot about this awesome PR

VincentLanglet:
> Tomorrow, thanks for pinging me, I forgot about this awesome PR

@rumpl Tomorrow was yesterday, did you need another ping ? ;)

rumpl:
As soin that conflicts are treated ans codecov is happy :)

rumpl:
@Urik do you think you will have the time to finish this PR?

azai91:
Any update on the status of this PR?

Urik:
@rumpl didn't realize about the coverage! 
I just moved to Montreal and have been super busy, but next week's weekend I'll find some time to check the tests.

rumpl:
Montreal, awesome! Take your time, whenever is good for you is good for me :) Just wanted to know if you still wanted to work on it still or not.

Keep warm, the winter over there can be quite cold I hear :)

Urik:
@rumpl @43081j 
Finally settled and motivated to do some at home stuff. I've rebased, added tests, and fixed formating. Believe it's good for review now.

rumpl:
Oooh, a christmas gift?! Thank you so much, will review quickly

src/utils/configuration-reader.js
```diff
@@ -0,0 +1,60 @@
...
+function parseCsvArray(value) {
+  return (value || '').split(',');
+}
+
+export function getCliArgs(args, version) {
+  return yargs(args)
+    .usage('Usage: $0 [DIRECTORY]')
+    .boolean(['dev', 'ignore-bin-package', 'skip-missing'])
+    .default({
+      dev: true,
```
rumpl: `dev` was deprecated and removed

README.md
```diff
@@ -82,6 +82,30 @@ All of the arguments are optional:
...
+***.depcheckrc***
+```
+ignores: ["eslint", "babel-*"]
+skip-missing: true
+```
+
+**Important:** if provided CLI arguments conflict with configuration file ones, the CLI ones will take precedence over the rc file ones.
+
+The rc configuration file can also contain the following extensions: `.json`, `.yaml`, `.yml`, `.js`, `.config.js`.
+If the `.js` or `.config.js` extension are used, then the file should export a Javascript object:
```
rumpl: Is there a way to tell cosmiconfig not to load javascript files? I really want the configuration to be static

src/utils/configuration-reader.js
```diff
@@ -0,0 +1,60 @@
...
+      'ignore-bin-package': false,
+      'skip-missing': false,
```
znarf: If we set a default in the CLI parameters here, we will never be able to set the values from the configuration file.

Fix: the default values of the CLI parameters should be removed, they're not necessary.

rumpl:
@Urik, @znarf made some pretty good points [in his comment on his PR](https://github.com/depcheck/depcheck/pull/491#issuecomment-569513164)

src/utils/configuration-reader.js
znarf: `defaultOptions` has the same purpose later and it's great.

test/configuration-reader.js
```diff
@@ -0,0 +1,111 @@
...
+      let thrownException = null;
+      try {
+        await configurationReaderModule.getConfiguration([], 'test', '1.0');
+      } catch (err) {
+        thrownException = err;
+      }
+      // eslint-disable-next-line no-unused-expressions
+      thrownException.should.not.be.null;
+      thrownException.should.be.instanceof(ConfigurationParsingException);
+      // const configResult = await configurationReaderModule.getConfiguration([], test, '1.0');
```
znarf: Why is that commented? Work in progress or candidate for deletion?

Urik: Leftover manual test, thanks for the catch.

README.md
Urik: Yeah, will be pushed with my next change.

Urik:
@rumpl @znarf I agree that's a good change, we should move `getParsers` from cli.js into the CLI configuration reader.
I'll have that tomorrow.

rumpl:
Everyone happy new year!

Urik:
Happy new years guys! 
I implemented the changes @znarf suggested for the `parsers` usage in the configuration file, removed the `default` calls in Yargs since we deal with that in `defaultOptions`, and removed support to `.js` config files.

rumpl:
Happy new year to you too @Urik !

rumpl:
🎉🎉🎉🎉🎉 THANK YOU 🎉🎉🎉🎉🎉🎉

Thank you so much @Urik for this PR!

krystofwoldrich:
Not sure if this is the right place to ask, but anyway, is there any support for `extends` functionallity? For example I would have base file .depcheckrc.base.json and then .depcheckrc.json which would extend the base file?

PR #415, opened 2019-08-27
Add support for integration tests jest configuration

idan-at:
Sometimes we have several test configurations (unit, it, e2e, contract), and we separate them by using different jest configs (usually we use different environments).

since depcheck only searches for `jest.conf(ig).js(on)` it does not understand that our used modules in `jest.it.config.js` are dev dependencies.
this should solve it.


idan-at:
@rumpl thanks for merging!

will you be able to publish a new version soon? :)

rumpl:
@idan-at No, thank you for contributing! 

I'll work on typescript support today, hopefully I'll have something working, if not I can make a new release for sure. 👍

rumpl:
@idan-at I just released version 0.8.4 with this change in it, enjoy! :)

PR #417, opened 2019-09-01
Make vue-template-compiler require more flexible

dword-design:
I have depcheck included in a 3rd party CLI tool that builds other projects. This means that vue-template-compiler is not included in the same project as depcheck. This PR makes the require of vue-template-compiler more flexible and also checks for existence in process.cwd().

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
     "builtin-modules": "^3.0.0",
     "deprecate": "^1.0.0",
     "deps-regex": "^0.1.4",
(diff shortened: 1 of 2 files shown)


rumpl:
Hey @dword-design,

your branch wasn't up to date with master, we are depending on vue-template-compiler now so no need to tryRequire or anything.

dword-design:
@rumpl yeah perfectly right, thanks!

PR #418, opened 2019-09-01
Make bin special compatible with yarn workspaces

dword-design:
Currently the implementation of the bin special does not work with yarn workspaces. The reason is that with yarn workspaces, modules can also be installed in a parent folder of the package folder.

I made the resolution logic more flexible. The only problem is that the packages need to have a main js file, otherwies the package cannot be resolved.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,9 +56,11 @@
     "lodash": "^4.17.11",
     "minimatch": "^3.0.2",
     "node-sass-tilde-importer": "^1.0.2",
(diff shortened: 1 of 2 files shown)


dword-design:
Hey, is this project actively maintained? I find it really useful :)

rumpl:
More or less actively maintained yes, when I have time, glad you find it useful :)

rumpl:
I'm closing this one since it was superseeded by #437 

Thanks for contributing

PR #419, opened 2019-09-12
Fix Gatsby config parser for plugins with options specified

gillisandrew:


gillisandrew:
Hey, to clarify it will only supports string literals forspecifying gatsby plugins. It would be nice to allow template literals (especially since they are quite common in the gatsby-config examples) but I'm not familiar enough with the babel AST to accomplish that. I've added a few tests but testing whether babel properly parses strings is a bit out of scope IMO.

That said I may have a look at adding template when I have some free time.

rumpl:
@gillisandrew will you still work on this?

Issue #359, opened 2019-06-26
False positive when only importing typescript interfaces (so no runtime code) from a package


Zjaaspoer:
### Bug Description
Only importing TypeScript interfaces (so no runtime code) provides false positives.

### Versions
 - `depcheck --version`: 0.8.1

### Extra info
It seems like the analysis is done on the compiled code and therefore won't know about the imported interfaces (as they are removed during compilation)?

rumpl:
Hey, thanks for reporting this, yes this is a known issue, the fix is to use the typescript parser and not babel for typescript files, will get to it as soon as I have some time.

I'll close this one since we already have a ticket for this #163 

Cheers

Zjaaspoer:
ah ok, thanks, will track that ticket!

PR #422, opened 2019-09-19
Use babel on typescript and support @types

### LINKED ISSUE #421: False alert: dep imported in d.ts file
My project contains `.js` files coupled with their typescript typedefs in `.d.ts` files. I'm running
```
depcheck --parsers="*.js:es6,*.d.ts:typescript"
```

I'd think this would work since the usages in the d.ts files are just normal es import statements. The only unusual thing about the contents of a d.ts file are that they contain the `declare` keyword.

conartist6:
Surprisingly easy, in fact! Much, much easier than trying to work with typescript's barely-supported traversal API.

Fixes #421, #359, #163

### DIFF
--- a/src/check.js
+++ b/src/check.js
@@ -7,6 +7,8 @@ import builtInModules from 'builtin-modules';
 import requirePackageName from 'require-package-name';
 import { readJSON } from './utils';
 import getNodes from './utils/parser';
+import { getAtTypesName } from './utils/typescript';
+import { availableParsers } from './constants';
 
 function isModule(dir) {
   try {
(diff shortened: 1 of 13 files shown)


src/check.js
```diff
@@ -71,6 +73,19 @@ function getDependencies(dir, filename, deps, parser, detectors) {
         .flatten()
         .uniq()
         .map(requirePackageName)
+        .thru(_dependencies => (
```
conartist6: I'm sure it must be possible to factor this more cleanly, though it's not clear to me what the best way of extracting an API to allow this kind of processing would be, or if that would really be worth the effort.

conartist6:
Also note that this falls short of covering all the ways that @types types can be used.
Still left for future work are:
-  reference comments (`/// <reference path="..." />`)
-  usages of globals declared in @types types (support seems unlikely)
-  usages in `tsconfig.json` `compilerOptions.types` field

rumpl:
Awesome, thank you @conartist6 !

PR #432, opened 2019-10-29
Add support for babel 7's scoped plugins

d-fischer:
Currently, you need to use the full preset or plugin name like `@babel/preset-env`.

This enables the short notation like `@babel/env` that already works similarly for older Babel versions (`babel-preset-env` => `env`).

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,10 +1,10 @@
 {
   "plugins": [
-    "babel-plugin-add-module-exports",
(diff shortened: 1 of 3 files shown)


.babelrc
```diff
@@ -1,10 +1,10 @@
 {
   "plugins": [
-    "babel-plugin-add-module-exports",
```
rumpl: Nice 👍

Issue #436, opened 2019-10-31
Node modules may be in parent directories


sveyret:
### Bug Description
When using the `bin` special parser, the binary module is only looked in current project `node_modules` folder. But actually, node may also search the module in parent folders. See https://nodejs.org/api/modules.html#modules_loading_from_node_modules_folders

This feature is especially mostly used by yarn workspaces or lerna monorepo.

### Versions
 - `node -v`: any
 - `npm -v`: any
 - `depcheck --version`: 0.8.4

### Extra info
I already did the correction in my system and will provide a PR. I've just seen that PR #418 seems to give a correction to it, but it is introducing new dependencies while my correction only adds a few lines to bin.js.

rumpl:
Closed by #437

PR #437, opened 2019-10-31
special/bin: climb up directories to find binary module

sveyret:
Issue #436


sveyret:
Note that the same problem exists for ignoreBinPackages. I am about to update it too, please wait before merging.

rumpl:
It would seem that you need to rebase your branch against master too

rumpl:
Could you add a test too ?

sveyret:
> It would seem that you need to rebase your branch against master too

Are you sure? It says it is already up to date…

rumpl:
Actually it fails because of the changes I made in a dependent build due to #433 

As soon as that build passes I'll merge it and then you can rebase.

Please don't ask what is this dependent build, I'll work on removing it :)

rumpl:
Some tests would be nice yes :)

Thank you for your work, it's appreciated

sveyret:
I added tests and also made some other modifications:

- changed the per-directory .gitignore into a single one into the test directory (they all were the same)
- removed a test which was duplicate

Each of these modification are in a separate commit for if you do not want to take them.

rumpl:
Thanks! I’ll take a look at it closely tomorrow but it looks good from my phone. Cheers

Also, happy halloween if you are into that kind of things :)

rumpl:
Unrelated but I used to work in Annecy, tres belle ville :)

sveyret:
@rumpl Just a reminder as you said you'd have a look at this one today… :smile:

test/spec.js
```diff
@@ -499,19 +499,6 @@ export default [
       },
     },
   },
-  {
```
rumpl: Why did you remove this test?

sveyret: If I understood well, it's doing exactly the same thing as 2 tests above (report unused bin dependencies if ignoreBinPackage equal false).

rumpl: ooooh good catch!

rumpl:
Thanks for all the work you have been doing lately!

sveyret:
You're welcome. Actually, I'm just making it work for my project… :wink:

rumpl:
Need anything else or do you want me to make a release just for you?

sveyret:
Well, I made everything I needed, so yes, if you can make a release, that'd be great!

rumpl:
Lets get #451 in it too

Issue #448, opened 2019-11-01
special/webpack: babel-loader presets are not detected


sveyret:
### Bug Description
When a babel preset is used in webpack configuration file, it is not detected as a dependency.

### Code snippets
Where the dependency is used:

```
module.export = {
    module: {
      rules: [
        {
          test: /\.js$/, use: {
            loader: 'babel-loader',
            options: { presets: [
              ['@babel/preset1', { option: 'option' }],
              '@babel/preset2'
            ]},
          }
        },
      ],
    },
}
```

Where the dependency is listed in package.json:

```
"devDependencies": {
  "@babel/preset1": "7.6.3",
  "@babel/preset2": "7.6.3",
}
```

### Versions
 - `node -v`: v8.12.0
 - `npm -v`: 6.4.1
 - `depcheck --version`: 0.8.4

### Extra info
Currently  creating a PR for that.

rumpl:
Closed with #449

Issue #446, opened 2019-11-01
special/webpack: entries should also be scanned


sveyret:
### Bug Description
In the `entry` section of webpack configuration, we may also specify dependencies which should be scanned. It is the case, for example, of the @babel/polyfill, of special fonts, etc.

### Code snippets
Where the dependency is used:

```
module.exports = {
  entry: [
    '@babel/polyfill',
    'typeface-roboto',
    './src/index.ts',
    './src/images/favicon.png'
  ]
}
```

Where the dependency is listed in package.json:

```
"dependencies": {
  "typeface-roboto": "0.0.75"
},
"devDependencies": {
  "@babel/polyfill": "7.6.0",
}
```

### Versions
 - `node -v`: v8.12.0
 - `npm -v`: 6.4.1
 - `depcheck --version`: 0.8.4

### Extra info
I am currently working on a PR for this issue.

rumpl:
Closed with #449

Issue #246, opened 2018-02-26
Support for Typescript webpack files


cbrianball:
It looks like the webpack special only works if the webpack config file is a .js file. I use .ts files for my webpack configuration.

lijunle:
Hi,

You are right. It is only recognize the .js file. It could be great that
you can help on it.

Brian Ball <notifications@github.com>于2018年2月27日 周二上午5:21写道：

sveyret:
FI, I'm currently working on this…

rumpl:
Closed with #449

sveyret:
Should we re-open this one because of the `ts-node` requisite? Or simply add a comment in any document?

rumpl:
Yeah, I'll reopen it, so that I don't search why:

The regex for webpack configuration was changed to accept .ts files but we can only require this kind of file if depcheck is used as a library with a ts-node runtime.

For this to trully work we need a way to be able to load .ts configuration files. This will surely mean that we need to depend on typescript directly, which I don't mind that much.

sveyret:
@rumpl did you receive my e-mail? I sent it to you GMail account from a personal server, and I know that GMail is sometimes touchy about personal servers, and so may put e-mails in spams.

rumpl:
Just saw it, the email I have on github is an email that I don't look every day, will respond today, have to clean up my appartment now :)

sveyret:
PR #459 now should fully correct the issue. @rumpl, I think it can be closed.

PR #449, opened 2019-11-01
special/webpack improvements

sveyret:
Correcting some issues about webpack special parser.
Tests included.
This is correcting issues #246, #446 and #448.

### DIFF
--- a/src/special/webpack.js
+++ b/src/special/webpack.js
@@ -44,6 +44,24 @@ function getLoaders(deps, loaders) {
     .value();
 }
 
+function getBabelPresets(deps, loaders) {
+  return lodash(loaders || [])
+    .filter((item) =>
+      typeof item !== 'string'
+      && item.loader
+      && item.loader === 'babel-loader'
+      && item.options
+      && item.options.presets
+      && Array.isArray(item.options.presets))
+    .map((item) => item.options.presets)
+    .flatten()
+    .map((preset) => (Array.isArray(preset) && preset.length > 0 ? preset[0] : preset))
+    .filter((preset) => typeof preset === 'string')
+    .intersection(deps)
+    .uniq()
+    .value();
+}
+
 function parseWebpack1(module, deps) {
   const loaders = getLoaders(deps, module.loaders);
   const preLoaders = getLoaders(deps, module.preLoaders);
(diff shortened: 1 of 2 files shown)


src/special/webpack.js
```diff
@@ -66,18 +84,45 @@ function parseWebpack2(module, deps) {
...
+    .intersection(deps)
+    .uniq()
+    .value();
 }
 
 export default function parseWebpack(content, filepath, deps) {
   const filename = path.basename(filepath);
   if (webpackConfigRegex.test(filename)) {
-    const module = require(filepath).module || {}; // eslint-disable-line global-require
+    const wpConfig = require(filepath); // eslint-disable-line global-require
```
rumpl: You changed the regex to include .ts config files, wouldn't this fail for a ts file?

sveyret: I am working with TypeScript, and it's working like a charm… :smile:

rumpl:
Beautiful, thanks!

src/special/webpack.js
sveyret: Actually, I've got a doubt. It's working with my project but I'm calling depcheck from a TypeScript file using ts-node. Maybe it'd be a requirement in order to use this feature.

rumpl: Ok so yeah it works for you because of ts-node. 

I think we should revert the ts loading until we find a good way of dealing with this

sveyret: I'll do some more tests directly using the command line.
But anyway, will it hurt to keep this commit? I mean, at worst, if people is not using it with the same requisites as I do, it will fail while reading the config file, and therefore, work as before. So it should be improved, but already can be a useful feature…

rumpl: The cli could fail for users where it didn't fail before. At least could you make a PR and make it so that it uses the `tryRequire`?

sveyret: So, after my tests, I can confirm that:

- when using `ts-node`, it works well, even with command line (I tested with `npx ts-node ~/depcheck/bin/depcheck.js --specials=bin,tslint,webpack`);
- when executing depcheck with `node`, the webpack configuration file is ignored, but the rest of the process is working correctly.

rumpl: Great then, nothing broke :)

PR #451, opened 2019-11-01
special/webpack: Use tryRequire to load configuration file

sveyret:
This will prevent the process to fail if loading a TypeScript configuration file when depcheck is not executed with ts-node.

Related to PR #449.


rumpl:
We already have a util [tryRequire](https://github.com/depcheck/depcheck/blob/69005fbeea5920fceaba15172321af57d6bb8be4/src/utils/index.js#L21), no need to add a new dependency

sveyret:
OK, I didn't see it. I will correct that…

PR #459, opened 2019-11-02
Load typescript transpiler if available

sveyret:
This will allow the require of the webpack configuration file, even in TypeScript, without the need to run depcheck with ts-node. This needs that depcheck can find the typescript dependency, either directly in the scanned project, in the current directory, or in the directory of depcheck.

Issue #246

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,7 +91,8 @@
     "patch-version": "^0.1.1",
     "prettier": "^1.18.2",
     "proxyquire": "^2.1.3",
-    "should": "^13.2.3"
+    "should": "^13.2.3",
+    "typescript": "^3.6.4"
   },
   "nyc": {
     "sourceMap": false,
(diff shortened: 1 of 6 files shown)


sveyret:
Adding the typescript dependency removed all the optional keys from the `package-lock.json` file. Don't know if it is a problem…

rumpl:
Not a problem no

VincentLanglet:
> Adding the typescript dependency removed all the optional keys from the `package-lock.json` file. Don't know if it is a problem…

I think this depend of the npm version you use

package.json
```diff
@@ -16,8 +16,8 @@
   "scripts": {
     "component": "node -r @babel/register ./build/component.js > ./dist/component.json",
     "compile": "babel src/ -d dist/",
-    "depcheck": "node ./bin/depcheck.js --ignore-dirs=fake_modules",
```
rumpl: We have a nifty [file for this](https://github.com/depcheck/depcheck/blob/master/false-alert.js), you can even add a comment as to why you are adding the import

PR #468, opened 2019-11-03
feat: detect when ran as a lifecycle event

### LINKED ISSUE #460: False positive depcheck
Depcheck is reporting himself as unused dependency.

rumpl:
When a command is ran with yarn (like yarn depcheck), we shouldn't
output it as an unused dependency.

Closes #460

### DIFF
--- a/test/special/npm-scripts.js
+++ b/test/special/npm-scripts.js
@@ -1,0 +1,30 @@
+/* global describe, it */
+
+import 'should';
+import parse from '../../src/special/npm-scirpts';
+
+describe('npm scripts special parser', () => {
+  it('should ignore when not a lifecylce event', () => {
+    const lifecycle = process.env.npm_lifecycle_event;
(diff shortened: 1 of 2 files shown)


src/special/npm-scirpts.js
```diff
@@ -0,0 +1,12 @@
+import * as path from 'path';
```
VincentLanglet: @rumpl You made a typo for the name of the file

rumpl: I'm tired... Thanks

test/special/npm-scripts.js
```diff
@@ -0,0 +1,30 @@
+/* global describe, it */
+
+import 'should';
+import parse from '../../src/special/npm-scirpts';
```
VincentLanglet: I think there is a type `scirpts`

PR #469, opened 2019-11-04
Add TTypeScript transformer special parser

sveyret:
Here is the special parser for ttypescript (https://github.com/cevek/ttypescript).


rumpl:
Nice, could you add it to the list of specials in the README.md?

PR #470, opened 2019-11-04
Automate building of TypeScript definitions

sveyret:
Because I almost forgot to update index.d.ts last time I added a special parser, I modified the component.js script in order to make it possible to rebuild this index.d.ts file.

### DIFF
--- a/index.d.tmpl
+++ b/index.d.tmpl
@@ -43,46 +43,7 @@ declare namespace depcheck {
       [filePath: string]: any;
     };
   }
-
-  const parser: {
-    coffee: Parser;
-    es6: Parser;
-    es7: Parser;
-    jsx: Parser;
-    sass: Parser;
-    typescript: Parser;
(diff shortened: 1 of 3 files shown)


index.d.tmpl
```diff
@@ -43,46 +43,7 @@ declare namespace depcheck {
       [filePath: string]: any;
     };
   }
-
```
rumpl: This is not norma is it?

sveyret: It is. Instead of directly writing the list of parsers, detectors and specials, the index is now a template, and the `component.js` script is completing it with all the files found in appropriate directories, just the same way it is creating the `component.json` file.
If you are specifically talking about the empty line, each group (parser, detector and special) is written starting with an empty line. If you run `npm run component`, you will see that the generated `dist/index.d.ts` is the same as the previous `index.d.ts` (at least, it should be!)

PR #473, opened 2019-11-08
Use node resolution for dependencies

sveyret:
I noticed that peer dependencies were not read on my monorepo project. That was because the modules were actually in a parent directory (the monorepo root). This was the same problem I already corrected with the bin special parser (see #437).
I created a `loadMetadata` function which is using the node resolution system, but based on the project root directory, and used it everywhere it seemed appropriate (including on my previous corrections regarding bin modules).


sveyret:
@rumpl 
Pleasure! :smile: 
Would that be possible to make a new release in the coming days?

rumpl:
Ask and you shall receive: https://github.com/depcheck/depcheck/releases/tag/0.9.1

PR #474, opened 2019-11-08
Support prettier shared configuration

sveyret:
It is possible to define a prettier entry in package.json in order to get a shared prettier configuration in an external module. This PR make depcheck see this dependency.

See https://prettier.io/docs/en/configuration.html#sharing-configurations

### DIFF
--- a/test/fake_modules/prettier/package.json
+++ b/test/fake_modules/prettier/package.json
@@ -1,0 +1,5 @@
+{
+  "name": "my-cool-library",
+  "version": "9000.0.1",
+  "prettier": "@company/prettier-config"
+}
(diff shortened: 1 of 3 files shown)


test/fake_modules/prettier/package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "my-cool-library",
+  "version": "9000.0.1",
```
rumpl: It’s over 9000!!!!

😂

sveyret: It's a copy of the given example on the prettier page… :wink:

PR #478, opened 2019-11-14
Try to load eslint.js modules without a module.exports wrapper

rjatkins:
* If the module already uses regular CommonJS syntax, we won't be able to parse it with another wrapper
 * Also fix the custom config case too
 * Also fix the result for parse to return null, to match what evaluate will return if we parse successfully, but nothing is found. This avoids breaking the tslint invalid custom parser case, but breaks the test for an invalid default config. I think this second test is the one that needed to change

### DIFF
--- a/src/utils/linters.js
+++ b/src/utils/linters.js
@@ -23,8 +23,14 @@ export function parse(content) {
     // not valid JavaScript code
   }
 
+  try {
+    return evaluate(content);
+  } catch (error) {
+    // not valid JavaScript code
+  }
+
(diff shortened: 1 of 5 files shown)


rumpl:
Hey @rjatkins I merge another PR and updated your branch but now a test is failing, could you take a look at what's going on? Thanks

rjatkins:
All fixed. Ready for you to re-review and merge, whenever you are.

rumpl:
Thanks!

PR #479, opened 2019-11-25
Document and provide types for the API's package option

edsrzf:
The `package` property allows specifying dependencies without parsing a `package.json`. I'm finding this useful in my code and thought it'd be nice to document it and make it easier to discover.


rumpl:
Thanks!

PR #483, opened 2019-12-12
Add possibility to pass a --config parameter with a config file

dword-design:
This PR adds a --config CLI parameter which allows to pass an external config file.


znarf:
@dword-design Seems this should now be closed or revisited since https://github.com/depcheck/depcheck/pull/408 is merged. What do you think?

dword-design:
@znarf I haven't tested the other PR yet but it seems like it also finishes this PR. So I'd say close and maybe revisit 👍.

PR #490, opened 2019-12-29
Better prettier configuration

znarf:
- add scripts to check the whole repository `npm run prettier` and `npm run prettier-check`
- apply with `npm run prettier -- --write`
- add `.prettierignore`
- ~~add `pre-commit` hook using `husky` and `lint-staged`~~
- add check to CI to prevent formatting regressions


rumpl:
Hey @znarf thank you for this, could we just remove the husky integration? I really am not a fan of pre-commit checks that take time, I want my `git` to be fast, checking this on the CI is ok, but not on my machine. We can add it later if we see that too many contributions from external people are failing because of the prettier check.

Cheers

rumpl:
Could you add the prettier check also to the [appveyor](https://github.com/depcheck/depcheck/blob/master/appveyor.yml) and [travis](https://github.com/depcheck/depcheck/blob/master/.travis.yml) files? 

We need less systems for our CI...

znarf:
@rumpl I personally have good experience with `pre-commit` hooks for `prettier`. I understand your concerns and will remove it.

Haven't noticed `appveyor` and `travis` yet, will look at it.

rumpl:
Thanks!

PR #491, opened 2019-12-29
Support for JSON configuration in package.json

znarf:
I want to be able to add a global configuration in projects adding a `depcheck` section in `package.json`. This is a work in progress to get feedback.

I just saw there was an ongoing similar attempt here https://github.com/depcheck/depcheck/pull/408, I have to check what is the overlap.

### DIFF
--- a/doc/pluggable-design.md
+++ b/doc/pluggable-design.md
@@ -71,7 +71,30 @@ var opts = {
       depcheck.parser.json1,
       depcheck.parser.json2,
     ]
+};
+```
+
+### Use Parser From JSON configuration
+
+With the JSON configuration (set in package.json), you can configure it like the following.
+
+```json
+"depcheck": {
+  "parsers": {
+    "*.js": "es6",
+    "*.jsx": "jsx",
   }
+}
+```
+
+The above JSON configuration is equivalent to the following API options:
+
+```js
+var opts = {
+  parsers: {
+    '*.js': depcheck.parser.es6,
+    '*.jsx': depcheck.parser.jsx,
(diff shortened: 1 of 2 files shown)


rumpl:
Seems like we have 3 PRs up now for this particular feature, will have to take a really close look at all of those :)

Thanks

znarf:
I've now checked https://github.com/depcheck/depcheck/pull/408 and it's great! It's already supporting configuration in `package.json` out of the box. Also, I was afraid that `yaml` would be mandatory but JSON is also supported, nice.

For parsers, I do prefer the syntax I was proposing here, it's more semantic and closer to the internal API. In the other PR it's:

```json
  "parsers": [
    "*.js:jsx",
    "*.jsx:jsx",
    "*.ts:typescript",
    "*.tsx:typescript"
  ],
```

While I was proposing the following, no string manipulation required:

```json
    "parsers": {
      "*.js": "jsx",
      "*.jsx": "jsx",
      "*.ts": "typescript",
      "*.tsx": "typescript"
    },
```

Other small difference, the other PR is using `ignore` (closer to the CLI) while I was proposing `ignoreMatches` (closer to the API).

rumpl:
@znarf both of those things would be great to have in the other PR, I'll link your comment there

Urik:
@rumpl @znarf I agree that's a good change, we should move `getParsers` from cli.js into the CLI configuration reader.
I'll have that done tomorrow.

znarf:
@Urik fine by me. Looking forward having your great PR ready and merged!

PR #496, opened 2020-01-02
Support for next.js webpack configuration

znarf:
Webpack configuration in next.js projects is only available in `next.config.js`.

It's a bit tricky to load:
- `next.config.js` needs first to be required
- to get the webpack configuration the `webpack` function (from `next.config.js`) needs to be executed
- a `webpack` object needs to be passed to the function. I decided here to use a fake instead of trying to pass a real `webpack` object.


rumpl:
Am gonna need some more time to read this one, thanks for all the work you are doing lately! :)

znarf:
Note that I rebased and the CI failure is random, can't restart it.

znarf:
@rumpl It's passing now.

Issue #497, opened 2020-01-03
Support .*ignore files


znarf:
I think it would be useful to implement support for ".*ignore" files as an alternative to the `ignore-dirs` option. See for example what was done in prettier: https://github.com/prettier/prettier/issues/2294 and https://github.com/prettier/prettier/pull/2412

Pros:

- Most projects already have a `.gitignore` file and this could be automatically used. I don't think anyone wants do `depcheck` any paths ignored in `.gitignore`. I've seen `depcheck` crashing because it was not ignoring some of these files (in my case, huge binary files >= 100MB), it was difficult to troubleshoot.
- The syntax used in `.gitignore` and similar `.ignore` files is more powerful than the current `ignore-dirs` API. It allows using `glob` while the `ignore-dirs` API is only expecting exact matches on single level directory names. https://github.com/depcheck/depcheck/blob/master/src/check.js#L144

Cons:
- It's confusing to introduce duplicate functionality with `ignore-dirs` and an incompatible syntax. There needs to be a sensible transition strategy.

rumpl:
+100 for an ingore file, we can have both for some time and print a warning to say that the flag will be deprecated (we already had a deprecated flag, when I removed it 2+ years later no one complained.

znarf:
@rumpl would be great to to look at my PR!

znarf:
Implemented.

PR #498, opened 2020-01-03
Support for .*ignore files

znarf:
Implement https://github.com/depcheck/depcheck/issues/497

- Introduce `ignore` dependency: https://www.npmjs.com/package/ignore
  - "ignore is a manager, filter and parser which implemented in pure JavaScript according to the .gitignore spec"
  - "ignore is used by eslint, gitbook and many others"
- Keep supporting `ignoreDirs` with a different implementation
  - `ignoreDirs` is now deprecated
  - while mostly retro-compatible, this is technically a breaking change
- Introduce `ignorePatterns`:
  - aims to replace ignoreDirs, use patterns instead of single level directory names
- Introduce `ignorePath`
  - Support any ignore file passed as `ignorePath` (ie: .eslintignore, .prettierignore)
  - If no `ignorePath`, try to use `.depcheckignore` or `.gitignore` in that order
- Switch file walker from `walkdir` to `readdirp`
  - simpler: ability to work recursively
  - ability to filter both files (`fileFilter`) and directories (`directoryFilter`)
  - easier to work with `ignore` package
- Start ignoring based on file extension:
  - images (.png, .gif, .jpg, .svg)
  - fonts (.woff, .woff, .eot, .ttf)
  - archives (.zip, .gz)
  - videos (.mp4)

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,11 @@ All of the arguments are optional:
 
 `--ignores`: A comma separated array containing package names to ignore. It can be glob expressions. Example, `--ignores="eslint,babel-*"`.
 
-`--ignore-dirs`: A comma separated array containing directory names to ignore. Example, `--ignore-dirs=dist,coverage`.
+`--ignore-dirs`: DEPRECATED, use ignore-patterns instead. A comma separated array containing directory names to ignore. Example, `--ignore-dirs=dist,coverage`.
+
+`--ignore-path`: Path to a file with patterns describing files to ignore. Files must match the .gitignore [spec](http://git-scm.com/docs/gitignore). Example, `--ignore-path=.eslintignore`.
+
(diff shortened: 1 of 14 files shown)


znarf:
This PR will need to be squashed at some point. It's almost ready in the current state. Unfortunately, one feature is broken: the ability to follow symlinks. No easy fix, switching back from `readdirp` to `walkdir` would lead to some inelegant code that I don't want to commit!

znarf:
Need to try re-implementing using https://github.com/jprichardson/node-klaw

znarf:
@rumpl 3 months in the making and it's finally ready!

rumpl:
Let's hope it will take me less than that to review it 😅

znarf:
@rumpl Long week-end coming. Let's try to get that merged! :-)

Issue #500, opened 2020-01-04
eslint: dependency wrongly mark as unused


znarf:
This is likely introduced by https://github.com/depcheck/depcheck/issues/476 and https://github.com/depcheck/depcheck/pull/477

Back in this previous issue, the contributor (@sveyret) wrote:

> when analyzing eslint configuration dependencies, it does not look at this dependencies' dependencies.

It should not, seems to me this is defeating the general behavior of `depcheck`. Any dependency that is explicitly used or imported in the project should be declared in `package.json`, even if it's already made available by another dependency. I just spent some time troubleshooting this in a project.

I had the following eslint config:

```
extends:
  - standard-jsx
  - standard-react
```

And the following lines in `package.json`:

```
    "eslint-config-standard": "^12.0.0",
    "eslint-config-standard-jsx": "^8.1.0",
    "eslint-config-standard-react": "^9.0.0",
```

And depcheck was outputing:

```
Unused devDependencies
* eslint-config-standard-jsx
```

Because `eslint-config-standard-jsx` is a dependency of `eslint-config-standard-react`, it's marked as 'unused'. I don't think that's right.

sveyret:
I agree if dependencies' dependencies are dev dependencies, not if they are direct dependencies. Let me explain my use case : I have a project in which I define the way all my company's projects should behave, using an eslint configuration. In this project, I include (as runtime dependencies) all the modules I need for the rules described in it.
In the dependent projects, I only want to add this project I defined, because it should contain everything needed. I don't want to modify all my dependent projects each time I add a rule which requires a new dependency.

znarf:
I get what you want, but seems to me this is not the way depcheck is working generally. Example: when using `lodash`, it's not because some of my dependencies are using `lodash` that I should not add `lodash` to my package.json.

In your use case, seems to me you should avoid having the sub-dependencies  being detected by depcheck. Let's say your config is `eslint-config-mycompany` and it's using internally `eslint-config-node` and `eslint-config-react`. In your project, you should only configure with `eslint-config-mycompany` and not have `eslint-config-node` and `eslint-config-react` mentioned anywhere. This way, depcheck will not be reporting it.

Would be great if you could point to a sample repository so we could see your use case more clearly.

sveyret:
You're right. I don't remember right now what was not working this way…

I think that the repository I'm talking about is public. You can find it there :

https://git.duniter.org/gecogvidanto/gecogvidanto

sveyret:
[...]

PR #506, opened 2020-01-08
Improve recursive work of eslint configuration check

sveyret:
The eslint special parser should not require a dependency which is used in an inherited configuration
if this dependency is already a dependency of this inherited configuration, but it should require
a dependency which is directly used, even if it is already a dependency's dependency.

Issue #500


znarf:
@sveyret I tested and looks like it's properly fixing the case I reported in https://github.com/depcheck/depcheck/issues/500.

PR #512, opened 2020-01-10
Add `debug` package and messages to the project

znarf:
`debug` is a well known library to add some debugging messages that are silent by default.

The way to get the messages is to opt-in using the `DEBUG` environment variable such as `DEBUG=depcheck:* depcheck`.

When working on the library I always add console.log in a lot of places and it would be useful to have them available all the time. Will also be useful to troubleshoot issues with users of the package.

Gathering feedback on that!


rumpl:
Yes, this can definitely be useful!

znarf:
Ready for review. It would be already useful for me to have that merged.

More messages can be added overtime.

PR #514, opened 2020-01-13
Add Istanbul special parser

### LINKED ISSUE #508: Feature: add special parser for Istanbul.js
This would be great to have a parser for `istanbul.js` because it is possible to extend a configuration from a foreign package, and this package is currently not detected as used by `depcheck`.

See https://github.com/istanbuljs/nyc

sveyret:
Special parser for Istanbul configuration file which looks if the configuration is inheriting from another package.

Fix #508

### DIFF
--- a/src/special/istanbul.js
+++ b/src/special/istanbul.js
@@ -1,0 +1,40 @@
+import path from 'path';
+import { parse } from '../utils/cli-tools';
+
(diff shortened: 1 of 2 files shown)


sveyret:
Hi @rumpl, I'm back to business! :wink: I've got a few questions:

- Would you have time to take the 3 PRs I've done these last days?
- When do you plan to release a new version?
- What do you think is still missing for v1?

PR #515, opened 2020-01-14
Review special documentation

sveyret:
* Re-order items in order not to forget any.
* Remove the “Grunt” entry (there is no grunt special).
* Add missing “Istanbul”, “Karma” and “Prettier” entries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,21 +40,23 @@ npm install -g depcheck typescript
 
 The _special_ component is used to recognize the dependencies that are not generally used in the above syntax files. The following scenarios are supported by specials:
 
-- `bin` - Dependencies used in npm commands, Travis scripts or other CI scripts
-- `eslint` - [ESLint](https://www.npmjs.com/package/eslint) configuration presets, parsers and plugins
-- `tslint` - [TSLint](https://www.npmjs.com/package/tslint) configuration presets, parsers and plugins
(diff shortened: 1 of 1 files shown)


sveyret:
Hi @rumpl and thank you for merging previous PRs. Sorry to bother you but you did not answer to my questions, especially “What do you think is still missing for v1?” If there are a few improvements, I may find time and help you…

rumpl:
Hey @sveyret

Sorry for not responding, didn't see the questions :(

What's missing for v1? 
I would really like for depcheck to have better tests and have zero false positives/negatives on starter projects like `create-react-app`, `gatsby`, `angular` and a couple of others, I think vue has a starter project too?

As for impovements, let's shoot for the moon:
* complete rewrite, the archictecture is not great and it's not easy to follow the code
* typescript (?)

One other improvement would be a better configuration file for depcheck, mainly, for each special/parser/whatever the user should be able to specify where the configuration files are. Most of the issues I see are because depcheck couldn't find the configuration (for example, we don't always find the configuration file for webpack which leads to a lot of false negatives).

But more realistically I think that once the current PRs are merged I will release a v1.

sveyret:
OK, all these are big tasks which I will not be able to work on now.

Regarding configuration file, I opened an issue for using `Cosmiconfig`. I think it would be a good idea to find a way to give parameters to special/parser…

Thank you.

PR #518, opened 2020-01-22
Support .js config files

dword-design:
Is there any reason not to support .js config files?

The support has been explicitly removed in https://github.com/depcheck/depcheck/pull/408. However, *.config.js files have lots of benefits (config merging, advanced control structures etc.) and are common in several packages like babel.

Also, this PR adds support do directly pass functions to parsers, specials and detectors, opening the door to external packages implementing parser, specials and detectors.


rumpl:
Hey @dword-design yes, there is a really good reason why I don't want .js configuration files: I strongly believe that configuration should be static, declarative. depcheck has been fighting .js configuration files from babel, webpack and such for years, it's just a huge PITA to try and safely extract configuration from a dynamic JavaScript file.

dword-design:
Hey @rumpl, thanks very much for your reply! hmmm can you maybe elaborate on which problems arose in the past regarding babel and webpack config files in conjunction with depcheck?

Regarding the restriction to only static content in configs in general I'd say that it kinda makes sense but should be left to the users of the library, and the users can then restrict their projects in their individual way.

rumpl:
Extracting configuration from a js file is not easy, we try to eval the configuration but it fails often, and then we can't know what plugins the user is using it its configuration and then we get false positives.

dword-design:
@rumpl I usually just `require` the config. This assumes that the config is somehow downcompiled to a node compatible environment. The case where the config is in some babel-dependent format should be handled by the caller of the library, e.g. by calling @babel/register, to make sure that require uses the plugins of the project.

PR #520, opened 2020-01-31
feat: Detect inline webpack loaders

### LINKED ISSUE #236: False positive when using inline Webpack loader
This tool returns a false positive when calling a script using an inline Webpack loader. For instance:

```js
require('script-loader!slick-carousel');
```

Here, `depcheck` considers `slick-carousel` as unused.

rumpl:
This was never a good idea and I wish webpack would drop support for this...

Fixes #236

### DIFF
--- a/src/detector/extract.js
+++ b/src/detector/extract.js
@@ -1,0 +1,10 @@
+// Not fond of default exports, disable here */
+/* eslint import/prefer-default-export: off */
+export function extractInlineWebpack(value) {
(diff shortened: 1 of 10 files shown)


rumpl:
@znarf PTAL

Could you test this on your project and tell me if this indeed does fix the issue for you? 

Thanks

znarf:
@rumpl Thank you, just tested and it works for my case!

PR #524, opened 2020-02-28
Feat: add special `angular` detector

rjlopezdev:
# Motivation
Angular projects can contains hidden dependencies over `angular.json` & `angular-cli.json` file.
This dependencies could be `scripts`, `styles` or `assets` from `node_modules` folder.

# Resolution

I code a simply detector for checking this file & return used dependencies.

# ⚠️  Help wanted 

I coded a skeleton for use cases I consider fits. Do you think that it fills properly? If agree, I'll implement it 😸

### DIFF
--- a/src/special/angular.js
+++ b/src/special/angular.js
@@ -5,7 +5,7 @@ const ANGULAR_FILES = ['angular-cli.json', 'angular.json'];
 const ANGULAR_JSON_KEYS = ['scripts', 'styles', 'assets'];
 
 const traverse = (obj, candidates) => {
-  obj.keys((k) => {
+  Object.keys(obj).forEach((k) => {
     if (obj[k] && typeof obj[k] === 'object') {
       if (ANGULAR_JSON_KEYS.includes(k)) {
         candidates.push(obj[k]);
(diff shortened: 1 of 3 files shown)


znarf:
@rjlopezdev Looks good. Could you give some sample repositories to try that?

Tests descriptions are looking good too. I'm not the main maintainer but I see no blocker for a merge once they're there.

I'm surprised coverage doesn't seems affected yet according to codecov. The goal is normally to keep the same level of coverage as today.

src/special/angular.js
```diff
@@ -0,0 +1,43 @@
...
+      }
+      traverse(obj[k], candidates);
+    }
+  });
+};
+
+export default async function parseAngular(filename) {
+  const basename = path.basename(filename);
+
+  if (ANGULAR_FILES.includes(basename)) {
```
LinusU, line 21: This function could potentially be a bit cleaner with an early return:

```js
  if (!ANGULAR_FILES.includes(basename)) return false;
```

rumpl:
@rjlopezdev thaks for the PR, code looks good, we would just need some tests :)

rjlopezdev:
Sorry for delay, I'm so busy. I will add some other cases for angular projects & test It ASAP ❤️

rumpl:
No rush ;) just wanted to know if you still plan on working on this. Thanks for the PR :)

rumpl:
Ping @rjlopezdev :)

Sorry to bother you, do you still want to do this? I can carry it if you don't have the time

Cheers

PR #532, opened 2020-04-10
Support resolve.alias in fakeConfig for next.js

znarf:
Small fix necessary for my own usage on https://github.com/opencollective/opencollective-frontend/  https://github.com/opencollective/opencollective-frontend/blob/master/next.config.js

### DIFF
--- a/src/special/webpack.js
+++ b/src/special/webpack.js
@@ -135,7 +135,7 @@ function loadNextWebpackConfig(filepath) {
     plugins: [],
     module: { rules: [] },
     optimization: { splitChunks: { cacheGroups: {} } },
-    resolve: { alias: () => {} },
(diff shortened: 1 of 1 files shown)


src/special/webpack.js
```diff
@@ -135,6 +135,7 @@ function loadNextWebpackConfig(filepath) {
     plugins: [],
     module: { rules: [] },
     optimization: { splitChunks: { cacheGroups: {} } },
+    resolve: { alias: () => {} },
```
LinusU, line 138: From the usage you linked it seems like `alias` should be an object, and not a function, maybe I'm missing something?

znarf: You're absolutely right, it should be an object. Fixed.

PR #534, opened 2020-05-10
Chore update deps

rumpl:


rumpl:
For some reason travis doesn't send the report, the build passed...

PR #535, opened 2020-05-10
chore: build on linux and windows on GH actions

rumpl:
Things I did:

* added windows runner on github actions
* made them required before merging
* added the badge in the readme file
* removed codecov from dependencies, use codecov action for this
* removed travis build (travis will be only used for deploying)
* removed appveyor

Things left to do:
* move the deployment over to github actions too


LinusU:
Neat! 👍 

Should probably change the `name: linux ci` also

rumpl:
@LinusU yeah will rename, ideally we wouldn't need tavis and appveyor any more, will do some tests.

rumpl:
I already tried gituhb actions on windows but they were flaky at the time, hopefully it's better now

rumpl:
Breaks because of symlinks :'(

.github/workflows/nodejs.yml
```diff
@@ -26,7 +35,12 @@ jobs:
           npm run lint
           npm run prettier-check
           npm run test-coverage
-          npm run test-dependent
```
LinusU, line 29: Doesn't seem like this is run from anywhere anymore, should we also remove the package.json script & dependency?

rumpl: Done 👍

PR #578, opened 2020-08-12
fix: now supports multi ignore in ignorePattern

YonatanKra:
Support the case in which we have multiple ignores:
Before:
`["!d", "!c"]` would accept `c` (becuase arr.some(isAllowed) would return true).
After:
`c` is ignored


rumpl:
Could you rebase now that I merged your other PR please? Thanks :)

YonatanKra:
Thank you for a very useful lib :)

rumpl:
https://github.com/depcheck/depcheck/releases/tag/1.2.0 ;)

PR #581, opened 2020-09-04
fix(sass-parser): ignore local import in scss 

YonatanKra:
This solves an issue with depcheck in scss.
Apparently SCSS is not parsed the same way as SASS. When parsing SCSS, local imports (like importing _mixins.scss) is not overridden by the importer.
This is fixed by adding another SCSS parser.


YonatanKra:
@rumpl 
I see that the main difference in coverage is try/catch that are not firing. I can remove them, since I think my fix also removes the need for them. Should I? Or is there a different trick?
Cheers,
Yonatan

rumpl (approved):

PR #586, opened 2020-10-15
Add option to run depcheck through npx

elrumordelaluz:
for simplicity and avoiding installing global packages that need future updates (`npx` gets always the last version).


rumpl (approved):
Thanks!

PR #595, opened 2020-11-08
Support parser patterns based on file paths

### LINKED ISSUE #589: Allow to define patterns against absolute file path (again)
Hi! I'm opening this issue because it seems that the last time this issue was solved, the solution got lost at some point.

- Original issue: #107
- Solution: #108 
- Line in change log: https://github.com/depcheck/depcheck/blob/62640856028eea8192d9c89952dc37258ece0611/CHANGELOG.md#070-beta1-2016-03-10

The solution introduced most notably this change:

https://github.com/depcheck/depcheck/blob/98ad446cb84971d9362de807c3b41d69c6a11658/src/check.js#L123-L126

But on `main` and `master`, the change has gone missing, and the code still checks against the basename of files, making it impossible to have different parsers on different directories:

https://github.com/depcheck/depcheck/blob/adb216659b9f4d3ffe7e34fbae88d56e142ad090/src/check.js#L97-L103

Could it be that this change was lost when 0.7.0 was released? The change was part of 0.7.0-beta, and it seems there were some chronology issues during that time: https://github.com/depcheck/depcheck/issues/304

rumpl:
Carry #108, it got lost somehow at one point

Fixes #589


rumpl:
Codecov decided to be a PITA today, merging anyway

PR #626, opened 2021-03-02
add dependency on vue

dfee:
@vue/compiler-sfc has a peer dependency on vue: https://github.com/vuejs/vue-next/blob/fbf2d68bbcde75c95363e1fd17946c720d0f3925/packages/compiler-sfc/package.json#L34


rumpl:
@dfee thanks for the PR, you need to use `npm add` so that the lockfile is updated too.

I wonder why our tests are green, we do check depcheck against depcheck. Did you stumble upon an error?

dfee:
I'm using [pnpm](https://github.com/pnpm/pnpm) as my package manager, and I have [`strict-peer-dependencies`](https://pnpm.js.org/en/3.0.0/pnpm-install#strict-peer-dependencies) enabled.

package-lock.json
```diff
@@ -1,9949 +1,8 @@
 {
   "name": "depcheck",
   "version": "0.0.1",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
rumpl, line 4: Looks like you are using some outdated npm version

rumpl:
Why would the compiler ask for vue as a peer dependency 🤔

rumpl:
So I just talked to the vue people on their discourse, `compiler-sfc` doesn't really need `vue` to be installed,  the `peerDependency` is there only to show a warning if your vue version differs with the compiler one.

So I'm not sure we do want to add vue as a dependency since it's not really needed...

Screenshot 2021-03-02 at 5 25 31 PM

package-lock.json
dfee: I just fixed that. I was using the one that ships w/ Node LTS (i.e. 14) – that is npm v6.

rumpl:
This will surely be fixed in vue 3.0.8, [as soon as this PR is merged](https://github.com/vuejs/vue-next/pull/3200).

I think we can close this one then

PR #629, opened 2021-03-18
Remove punycode from the buildin modules

### LINKED ISSUE #628: punycode being listed as unused, although the built in version is deprecated
### Bug Description

If the project includes `punycode` as a dependency in `package.json`, it gets listed as unused even when it's not.

This is because `punycode` is also included as a builtin in the Node API, but as it's been deprecated there since v0.7 it should now be included as an explicit dependency.

In fact, given that's it deprecated in the NodeJS API and has been for a long time, it might be helpful to list it as a missing dependency if it's used but hasn't been explicitly declared as a dependency in `package.json`.

See https://nodejs.org/api/punycode.html#punycode_punycode

rumpl:
Fixes #628


jdrouet (approved):
LGTM

PR #631, opened 2021-04-29
Update config reader to work in mono repo

AminFazlMondo:
The config reader is aware of the context of the directory that the package.json file is in and will search only the same directory.
This is useful in the context of monorepo where you want to  run this with multiple different package.json file and their individula version of config file.

### DIFF
--- a/src/utils/configuration-reader.js
+++ b/src/utils/configuration-reader.js
@@ -66,7 +66,7 @@ function returnNull() {
   return null;
 }
 
-export async function getRCFileConfiguration(moduleName, filename) {
+export async function getRCFileConfiguration(moduleName, filename, dir) {
   try {
     const configFileExplorer = cosmiconfig(moduleName, {
       // this prevents cosmiconfig from picking up .js configuration files. "null" means no file was found.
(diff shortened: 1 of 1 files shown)


AminFazlMondo:
Thanks for the response @rumpl ,
I have made changes to fix linting issue of the pull request, hopefully should be passing now.

In regards to the context of the change, I have tried to use this useful package in our monorepos using [lerna](https://github.com/lerna/lerna).

The package structure is
`
my-lerna-repo/
  package.json
  .depcheckrc.yaml
  packages/
    package-1/
      package.json
      .depcheckrc.yaml
    package-2/
      package.json
      .depcheckrc.yaml
`
As you can see, each of the packages has its own config file for depcheck, and we want to be able to run the same command without needing to change the working directory (ie, `lerna run depcheck`)

Hopefully that clarifies the reason behind this change, please let me know if you need more details

> @AminFazlMondo  sorry about that, I just started the CI and it seems red, could you take a look at it.
> 
> Also, I'm not sure I 100% understand what issue this fixes, could you show me some examples please?
> 
> Cheers

PR #638, opened 2021-05-20
feat: mark certain deps as optional

maraisr:
Certain people dont actually use `scss` or `vue` —so dont force those people to have it installed. Mainly solving the vue problem where `@vue/compiler-sfc` has `vue` marked as a peer, so complains on install. This way we make it optional.


rumpl:
Vue problem [should be fixed soon](https://github.com/depcheck/depcheck/pull/626), I didn't have the time to check this yet but will soon

maraisr:
Friendly bump @rumpl

rumpl:
Hey @maraisr 

I really don't like peerDependencies, also, this tool is supposed to be installable as a cli, how would that work with peer dependencies?

I know we have a lot of dependencies but that's the price of handling a lot of tools/frameworks that all have their own way of defining their configuration

PR #641, opened 2021-05-22
Respect overrides in eslint configs

### LINKED ISSUE #570: False positive: eslint overrides
In addition to including plugins in eslintConfig directly, you can also specifically require them in overrides. For example, in the `overrides` for tests, we include `eslint-plugin-mocha`.

This should be as easy as running the same checks for plugins on the elements of the `overrides` array as well as in the root config.

Xyaren:
Fixes #570 

Now iterates all override configs additionally to the main config.

### DIFF
--- a/src/special/eslint.js
+++ b/src/special/eslint.js
@@ -107,12 +107,26 @@ function resolvePresetPackage(preset, rootDir) {
 }
 
 function checkConfig(config, rootDir, includedDeps = []) {
-  const parser = wrapToArray(config.parser);
-  const plugins = wrapToArray(config.plugins).map((plugin) =>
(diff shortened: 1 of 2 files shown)


Xyaren:
I think I need help with that CI failure. I ran the tests locally and they were fine. What is happening?

Xyaren:
I tried to replicate the issue locally but I am not able to reproduce it. 
Developing on a Windows machine is not really helpful in that way.
2 Completely unrelated test seem to fail and eslint/prettier is complaining about my line separators, that git changed while checking out the repo...

rumpl (approved):
Thanks!

PR #652, opened 2021-06-08
Svelte

### LINKED ISSUE #650: Svelte Support
### Feature request description

It would be nice to have support for svelte projects (raw Svelte, Sapper or SvelteKit projects), similar to the React and Vue support that exists currently.

### Any extra info

Currently when you run `depcheck` on a Svelte project, it reports imports used within `.svelte` files as unused. Have not tested with raw Svelte projects or SvelteKit, but assuming that it has problems interpreting the `.svelte` files.

pouya-eghbali:
Adds Svelte support, closes #650 

Had to upgrade `node-sass` to `6.0.0` as I use Node 16.x for development.
Added Svelte to `devDependencies`
Uses dynamic imports for Svelte to make it an optional dependency.

A Svelte test case is added.
The following command runs without issues:

```bash
npm run prettier-check && npm run lint && npm run test
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -56,12 +56,13 @@
         "eslint-plugin-react": "^7.21.5",
         "fs-extra": "^9.0.1",
         "mocha": "^8.2.1",
-        "node-sass": "^5.0.0",
+        "node-sass": "^6.0.0",
         "nyc": "^15.1.0",
         "patch-version": "^0.1.1",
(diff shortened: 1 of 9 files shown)


rumpl:
This is beautiful!

rumpl (approved):
🎉

rumpl:
Guess we should add node 16 to our matrix

Issue #461, opened 2019-11-02
Packages used inside package.json are not detected


ktalebian:
I'm using [husky](https://github.com/typicode/husky), but when I run `depcheck` it doesn't recognize `husky` and labels it as unused:

```bash
$ depcheck
Unused dependencies
* husky
```

Same goes for `ts-node` that I use as `"test": "nyc mocha -r ts-node/register $(find test -name '*.ts')",`

`depcheck` does not complain about `nyc` being unused but also does not detect it as missing if I do not install it. Looks like packages that are used inside `package.json` file either as a hook or as part of a script are not recognized.

rumpl:
Hey @ktalebian 

Husky detection landed today! https://github.com/depcheck/depcheck/pull/453

It will be available in the next release, sometimes next week.

rumpl:
Could you open a new issue for ts-node, nyc please?

PR #657, opened 2021-07-23
feat: detect deps used with -r in package.json

Dylanlan:
I noticed depcheck wasn't detecting ts-node being used in one of my package.json scripts, such as:

`"test": "mocha -r ts-node/register test/**/*.ts"`

Looked like it was already detecting `--require`, but not `-r`, so I figured it made sense to add it. I found it was briefly mentioned in this issue too: https://github.com/depcheck/depcheck/issues/461


rumpl (approved):
Awesome!

Issue #633, opened 2021-04-30
[false positives] SCSS variables are tracked as packages, while they are not


theKashey:
### Bug Description

In some circumstances _@including_ a mixin is treated as an import

### Code snippets

Where the dependency is used:

```scss
@use "~@ui/styles/theme";
@use "~@ui/styles/mixin";

.icon {
  span {
    @include mixin.visuallyHidden;
    // mixin is missing in project dependencies
  }
}
```

### Extra info

The [current code](https://github.com/depcheck/depcheck/blob/main/src/parser/sass.js#L58) basically reacts at any "@at" rules.
I've tried to filter only `import`/`use` and it seems to work fine.

I can draft a PR if such solution is acceptable.

theKashey:
Technically speaking the required amount of parsing can be achieved via simple regexp - there is no real need to build full ast, solving #632 as well.

theKashey:
Still valid issue

ptmkenny:
I'm only seeing a potentially related issue with media queries.

**styles.scss**

```
@media only screen
  and (min-device-width: 375px) {
  .cropContainer {
    height: 250px;
  }
}
```

And `depcheck` reports:

`Missing dependencies
* only: ./styles.scss`

Issue #653, opened 2021-06-17
"Missing dependencies" local .scss file


nichoth:
### Bug Description

I have added to this project -- https://github.com/nichoth/mars-roving/blob/main/package.json#L20

Using like this
```
depcheck --ignores=\"cypress,esmify,uglifyify,depcheck,eslint,eslint-config-minlint,public/*\"
```

And it returns incorrectly that it's missing `./src/style/main.scss`, a local .scss file.

```
$ npm run deps

> @nichoth/mars-roving@0.0.0 deps
> depcheck --ignores="cypress,esmify,uglifyify,depcheck,eslint,eslint-config-minlint,public/*"

Missing dependencies
* %: ./src/style/main.scss
```

### Versions

- `node -v`:
`v14.3.0`
- `npm -v`:
`7.12.1`
- `depcheck --version`:
`1.4.1`

Issue #627, opened 2021-03-17
Feature Request: Support for plugins listed in netlify.toml files


princefishthrower:
### Feature request description

Awesome tool! Just discovering this tool today, man I had some dependencies from years ago... sheesh

Anyway, I would really like it if plugins in netlify.toml files were picked up. Currently they aren't, and listed incorrectly as unused.

### Code snippets

How the dependency is used (if applicable):

In a `netlify.toml` file, like so

```toml
[[plugins]]
package = 'netlify-plugin-contextual-env'
  [plugins.inputs]
  mode = 'suffix'
```

How the dependency is listed in package.json (if applicable):

```
"devDependencies": {
    ...
    "netlify-plugin-contextual-env": "^0.3.0",
}
```

### Any extra info

I could potentially create a pull request for this feature - I haven't looked at how this library works, if it's regex-based or similar I can't imagine this would be too hard.

rumpl:
Hey @princefishthrower 

Thank you so much for the kind message and I'm glad you find it useful! :)

If you could open a PR for netlify that would be awesome! 

What you need to do is add a new `special` that would parse the `netlify.toml` file and return the dependencies that are used in that file. In your example the special should read the toml and return the `netlify-plugin-contextual-env`. 

You can take a [look at my PR where I add the husky special](https://github.com/depcheck/depcheck/pull/453/files) to see how this works.

Feel free to ping me if you need help or if you think you won't have the time to implement this, we have the discussions activated on this repository.

Cheers

princefishthrower:
@rumpl - I may have spoke too soon here, it may be that NPM modules for netlify plugins may not be required... though I'm not sure and it may be a bug in the specific plugin. I will look into it.

As an aside, I notice my SCSS modules using the 'import' directive are marked as 'missing dependencies' - is there a setting I'm missing here or is this expected behavior for `depcheck`?

rumpl:
SCSS should work, if you can create a small repo that reproduces the sass error I could take a look at it.

princefishthrower:
@rumpl - sorry for my delay (and I'm still looking into the details here with netlify) anyway, here's an example of what I mean with the SCSS issues

I notice the problems lie mostly with animations and keyframes:

<img width="577" alt="Screen Shot 2021-03-26 at 11 10 19" src="https://user-images.githubusercontent.com/11366744/112616420-004dd600-8e24-11eb-95da-f79537298858.png">

[...]

PR #658, opened 2021-07-31
Fix scss parsing for non-import '@' keywords

Dylanlan:
It looked like the scss parser was treating everything with `@` as an import.

I'm not super familiar with scss, but I consulted this doc for what can be an import: https://sass-lang.com/documentation/at-rules

It looks to me that only `@use`, `@import`, and `@forward` should be treated as imports.

Relates to https://github.com/depcheck/depcheck/issues/653
And https://github.com/depcheck/depcheck/issues/627#issuecomment-808093756

### DIFF
--- a/test/fake_modules/sass/scss2.scss
+++ b/test/fake_modules/sass/scss2.scss
@@ -1,7 +1,20 @@
 @use "@test-dep/aFile2";
 @use "mixin"; // should ignore...
 @use "sass:math"; // should show only sass...
+@import "@test-dep/aFile3";
+@forward "@test-dep/aFile4";
+
+%ignore-extend {
+  margin: 0;
+}
 
 div {
+  @extend %ignore-extend;
   font-size: $font-size;
+}
+
+@keyframes ignore-keyframes {
+  70% {
+    margin-left: 90%;
+  }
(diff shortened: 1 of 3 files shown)


test/fake_modules/sass/scss2.scss
```diff
@@ -1,7 +1,20 @@
...
 div {
+  @extend %ignore-extend;
   font-size: $font-size;
 }
+
+@keyframes ignore-keyframes {
+  70% {
+    margin-left: 90%;
+  }
+}
```
rumpl, line 20: I'm not sure I understand why this was added? Is this needed for the tests?

Dylanlan: yeah I figured it would be useful to validate that `ignore-keyframes` is not treated as an imported dependency. I just gave it some fake properties, lemme know if you'd prefer it empty like `{}` or something

rumpl: Ah! Ok understood, we can leave it then

Thanks!

rumpl (approved):
LGTM

theKashey:
- fixes #633

PR #659, opened 2021-08-04
Ignore .idea/ 

DiegoRBaquero:
They were added in latest version: https://app.renovatebot.com/package-diff?name=depcheck&from=1.4.1&to=1.4.2

Such files shouldn't be published to npm

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -9,3 +9,4 @@ test/
 build/
 coverage/
 .nyc_output
+.idea/


rumpl:
Thanks @DiegoRBaquero 

Indeed, I published the last version manually because Travis no longer works and I have a lingering `.idea` folder, not sure why...

I'm little on the fence on this, if we start adding different folders from IDEs then there is no end on this. The `.idea` is not even in git so I don't really see why we would need this in the `npmignore` file.

The best thing we should do is to make it so that a github action makes the release.

DiegoRBaquero:
Understood, let me know if I can help

PR #660, opened 2021-08-04
Support next.config.js with next-transpile-modules

crossjs:
See: https://github.com/martpie/next-transpile-modules/blob/27ca78faec/src/next-transpile-modules.js#L345-L348


rumpl (approved):

rumpl:
Thanks!

PR #670, opened 2021-09-05
fix: allow using @types-module for typescript type-only imports

### LINKED ISSUE #526: Typescript dependencies with format @types/xxxx are incorrectly reported when using type imports
I have `@types/estree` in my `package.json` and a `.ts` which does:

```typescript
import ESTree from 'estree';
```
This is the correct syntax, since `tsc` complains if you import from the `@types` scope.

In this case, `depcheck` complains about missing dependency `estree`.

### LINKED ISSUE #568: False positive: types-only dependencies
Specific example: `@types/aws-lambda` is included to ease writing AWS lambda handlers, although `aws-lambda` is *not* included. The import is also specifically an `import type` in the code to further clarify that this is a type-only import.

This is showing up as `aws-lambda` being a missing dependency when it actually isn't.

rexxars:
Fixes #568
Fixes #526

I'm not sure if this is the best approach, given none of the other detectors seems to be using the second `deps` argument, but the alternative would be to change the returned value from an array of strings to an object that includes dependency names and an additional property for the _type_ of import, or similar.

### DIFF
--- a/src/detector/importDeclaration.js
+++ b/src/detector/importDeclaration.js
@@ -1,7 +1,18 @@
 import { extractInlineWebpack } from './extract';
 
-export default function detectImportDeclaration(node) {
-  return node.type === 'ImportDeclaration' && node.source && node.source.value
-    ? extractInlineWebpack(node.source.value)
-    : [];
+export default function detectImportDeclaration(node, deps) {
+  if (node.type !== 'ImportDeclaration' || !node.source || !node.source.value) {
+    return [];
+  }
+
+  // Typescript "import type X from 'foo'" - doesn't need to depend on the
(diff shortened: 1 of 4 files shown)


LinusU (approved):
Seems like an easy enough fix 👍

rumpl (approved):
Code looks good to me, sorry for the late review...

rumpl:
@rexxars thanks for the PR, could you amend your commit and force push to kickstart the CI please, I have no idea why I can run the actions on your PR...

rexxars:
Sure. Rebased.

rumpl:
Thanks @rexxars

rexxars:
Thanks for merging! Any chance of seeing this one in a release anytime soon?

PR #677, opened 2021-09-29
[WIP] generate index.js for components at compile-time

### LINKED ISSUE #676: API is not usuable with bundler
### Bug Description

https://github.com/depcheck/depcheck/blob/1c6efe2851b188d2c6bb24c5477c15692a1da2ed/src/constants.js#L5-L13

Please don't use dynamic requires here. Can we make them at compile time? My bundler isn't so smart, so it can't detect (and bundle) these imports. Because it isn't surrounded with try/catch, bundled code just crashes at runtime.

I suggest you to dynamically create `dist/parser/index.js` and `dist/detector/index.js` with all required imports, and in the code above just use namespace import.

## Steps to Reproduce

- Try to bundle code (I'm using esbuild for example)

### Versions

- `node -v`: v15.6.0
- `npm -v`: 7.4.0
- `depcheck --version`: 1.4.2

zardoy:
closes #676

I've rewrited `build/component.js` (because with console.logs the script needs to be executed 5 times, current implementation looks more clean imho) so it can now generate index files directly into `dist`. Source code looks something like this:
```ts
export { default as "bar" } from './bar';
export { default as "feross-standard" } from './feross-standard';
```

But now I understand that the tests runs against source code (not `dist`) and they are of course failing.

I'm not sure what should I do next, I can create another script and write indexes into `src/<component>` before tests run and gitignore them.


rumpl:
@zardoy thanks for the PR, please give me a couple of days to take a look at this, I'm sure I'll have some time this weekend.

PR #680, opened 2021-11-09
feat(types): improve typings

shian15810:
Improve typings for TypeScript users:

- Add missing properties to `Options`
- Create `Config` to match config file


rumpl:
Cool thanks!

PR #681, opened 2021-11-09
feat(eslint): add support for .eslintrc.cjs

shian15810:
As per [ESLint docs on `Configuration File Formats`](https://eslint.org/docs/user-guide/configuring/configuration-files#configuration-file-formats), `.eslintrc.cjs` is a supported file format of its kind, this PR adds support to it.

### DIFF
--- a/src/special/eslint.js
+++ b/src/special/eslint.js
@@ -171,7 +171,7 @@ function checkConfig(config, rootDir, includedDeps = []) {
   return result;
 }
 
-const configNameRegex = /^\.eslintrc(\.(json|js|yml|yaml))?$/;
+const configNameRegex = /^\.eslintrc(\.(json|js|cjs|yml|yaml))?$/;
 
 export default async function parseESLint(filename, deps, rootDir) {
   const config = await loadConfig('eslint', configNameRegex, filename, rootDir);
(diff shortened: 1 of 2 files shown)


rumpl:
Thanks! Quick question, what's a `cjs` and how is it different from a `js` file? Is there something special we need to do about a `cjs` file when loading it?

shian15810:
Currently Node.js support 2 types of module resolutions, namely CommonJS `require` and ECMAScript `import`.

Node.js consider a project to be of ECMAScript is there is a top-level `"type": "module"` in `package.json`, and will use module resolution logic of ECMAScript, else CommonJS is assumed

However ESLint support CommonJS-style `require` only, and will fail to read `.eslintrc.js` since module resolution logic of ECMAScript is assumed by NodeJs. Hence for such situation, one can rename the file to `.eslintrc.cjs` to let NodeJS know to use CommonJS-style `require` to read this file.

rumpl:
Thanks for the explanation, it's crystal clear 👍

shian15810:
Thanks for the quick actions! Is it possible to release a new version? Thanks in advance!!!

rumpl:
Will do, I just want to merge #670 and #679 before publishing a new version

rumpl:
@shian15810 I merged both of the PRs, new version coming tomorrow 👍

shian15810:
Bump

PR #682, opened 2021-11-17
feat(jest): add support for jest.config.cjs

shian15810:
As per [Jest docs on Configuring Jest](https://jestjs.io/docs/configuration), `jest.config.cjs` is a supported file format of its kind, this PR adds support to it.

### DIFF
--- a/src/special/jest.js
+++ b/src/special/jest.js
@@ -4,7 +4,7 @@ import { getContent } from '../utils/file';
 
 const _ = lodash;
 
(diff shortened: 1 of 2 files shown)


mdaverde:
It would be nice to have generic support for [cjs](https://github.com/depcheck/depcheck/issues/714) and [mjs](https://github.com/depcheck/depcheck/issues/733) to support more of the ecosystem

Issue #688, opened 2021-12-14
Running against .ts files fails with error: 'TypeError: p is not iterable'


danielreidwoebot:
### Bug Description
I'm trying to run depcheck against TypeScript code but it's failing with the error: ` TypeError: p is not iterable`. I've tried installing with `npm i -g depcheck` and also `npm i -g depcheck typescript`.

The error was generated after running:
```
depcheck path/to/my/repo --json
```

I have tried:
```
depcheck path/to/my/repo --parsers="*.ts" --json 
```
However, when I run this (^) command I get the error: `Cannot read property 'split' of undefined`. It does seem to be working as expected for .js files though.

Here's the stack trace for the error in the issue title.
```
"TypeError: p is not iterable\n    at /Users/danielreid/.nvm/versions/node/v14.18.2/lib/node_modules/depcheck/node_modules/@babel/parser/lib/index.js:10658:43\n    at Array.some (<anonymous>)\n    at hasPlugin (/Users/danielreid/.nvm/versions/node/v14.18.2/lib/node_modules/depcheck/node_modules/@babel/parser/lib/index.js:10654:18)\n    at validatePlugins (/Users/danielreid/.nvm/versions/node/v14.18.2/lib/node_modules/depcheck/node_modules/@babel/parser/lib/index.js:10693:7)\n    at getParser (/Users/danielreid/.nvm/versions/node/v14.18.2/lib/node_modules/depcheck/node_modules/@babel/parser/lib/index.js:15283:5)\n    at parse (/Users/danielreid/.nvm/versions/node/v14.18.2/lib/node_modules/depcheck/node_modules/@babel/parser/lib/index.js:15254:12)\n    at parseTypescript (/Users/danielreid/.nvm/versions/node/v14.18.2/lib/node_modules/depcheck/dist/parser/typescript.js:17:28)\n    at async getDependencies (/Users/danielreid/.nvm/versions/node/v14.18.2/lib/node_modules/depcheck/dist/check.js:71:18)"
```

### Versions

- `node -v`: v14.18.2
- `npm -v`: 6.14.15
- `depcheck --version`: 1.4.2

rumpl:
Can you try with `depcheck path/to/my/repo --parsers="**/*.ts:ts" --json`

danielreidwoebot:
> depcheck path/to/my/repo --parsers="**/*.ts:ts" --json

[...]

Issue #687, opened 2021-12-14
Depcheck started returning incorrect unused dependencies without any changes


tm-slavik57:
### Bug Description

I'm not sure it's related, but since of today depcheck started reporting incorrect unused dependencies.

### Versions

- `node -v`: 14.16.1
- `npm -v`: 6.14.12
- `depcheck --version`: 1.4.2

### Extra info

We are using `npx depcheck` and until yesterday it worked in our CI, starting from today it stopped working.
We didn't change anything, and on the same commit, it worked yesterday it stopped working today.

I tried to figure out what's wrong so I ran the depcheck locally on my laptop and it worked.
I then looked at the `depcheck` installed globally on my laptop and saw it's from July.
At this point the depcheck version is `1.4.2`

Running `npx depcheck` resulted in this output:
```
$ npx depcheck --config=./ci/.depcheckrc
No depcheck issue
```

After that I reinstalled it globally using:
```
$ npm install -g depcheck
C:\Users\Slavas\AppData\Roaming\npm\depcheck -> C:\Users\Slavas\AppData\Roaming\npm\node_modules\depcheck\bin\depcheck.js
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@~2.3.2 (node_modules\depcheck\node_modules\chokidar\node_modules\fsevents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@2.3.2: wanted {"os":"darwin","arch":"any"} (current: {"os":"win32","arch":"x64"})
```
The depcheck version didn't change and remained `1.4.2`

Now running the check command fails:
```
$ npx depcheck --config=./ci/.depcheckrc
Unused dependencies
* aws-sdk
* express
Unused devDependencies
* @types/chai
* @types/kafkajs
* chai
* chai-as-promised
* http-status-codes
* kafkajs
```

My assumption is when running `npx` is that it used the cached version that was on my laptop and after reinstalling it, it downloaded a new version and stopped working.

The thing I don't understand is that `depcheck -v` returned `1.4.2` both while the behaviour changed.

filipw01:
It's not just you, `@babel/parser` version 7.16.5 broke it, released 12 hours ago
Workaround, is to lock it using yarn resolutions
```
  "resolutions": {
    "depcheck/@babel/parser": "7.16.4"
  },
```

tm-slavik57:
[...]

PR #690, opened 2021-12-15
fix: arguments to @babel/parser

hedgepigdaniel:
Fixes https://github.com/depcheck/depcheck/issues/688

### DIFF
--- a/src/parser/typescript.js
+++ b/src/parser/typescript.js
@@ -31,7 +31,7 @@ export default async function parseTypescript(filename) {
       'objectRestSpread',
       'optionalCatchBinding',
       'optionalChaining',
-      { pipelineOperator: { proposal: 'minimal' } },
+      [ 'pipelineOperator', { proposal: 'minimal' } ],
       'throwExpressions',
     ],
   });


rumpl:
Thanks for the PR, if this fixes I will need some time to take a look and understand why this suddenly stopped working, [there is this other issue](https://github.com/depcheck/depcheck/issues/687) saying that an update to `@babel/parser` made things break for us. But I'm not sure I understand why we pick that updated babel parser, we are publishing a bundled package...

m4eryk:
@hedgepigdaniel can you update OR? it's so important

znarf:
@rumpl depcheck doesn't bundle a fixed dependency as it's using a loose version `"@babel/parser": "^7.12.5",` in package.json. If there is suitable version that matches, it will use it. That's valid for `@babel/parser@7.16.7`.

The breaking change in babel can be seen here: https://github.com/babel/babel/commit/f4236f43a119da17904c466e5d9c375141cb12a7

Seems the syntax in typescript.js was wrong but simply ignored before 7.16.7. It started to break after this.

Since the syntax was already correct in es7.js, jsx.js and vue.js, I think the change is safe.

znarf:
The change looks good but ultimately closing as there is a duplicate PR here #692

PR #691, opened 2021-12-17
fix: configure TypeScript correctly

judofyr:
The `pipelineOperator` plugin was configured incorrectly. This didn't cause a problem with @babel/parser v7.12, but it does cause v7.16 to crash when trying to parse code.

I've also updated `package-lock.json` (to start using the latest version of @babel/parser), but maybe it's better to remove the whole file so that the CI always uses the latest version of the dependencies in `package.json`?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,6 +5,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "depcheck",
       "version": "0.0.1",
       "license": "MIT",
       "dependencies": {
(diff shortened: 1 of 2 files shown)


YuniorGlez:
Checked in typescript node application and works like a charm ! There is a problem with latest babel versions and this PR solve the problem. Thanks @judofyr

pascalbayer:
I can confirm that the fix resolves the current issue with @babel/parser v7.16.6. Would be great to see the fix being merged soon. Thx @judofyr

YuniorGlez:
I know is not the best solution but i found a npm package called patch-package that you can use provisionally until this PR will be merged :)

znarf:
Looks good but closing as there is a duplicate PR here https://github.com/depcheck/depcheck/pull/692

PR #696, opened 2022-01-10
Remove the lockfile since it's not used on npm or with npx

rumpl:


rumpl:
Oh wow, this is going to be painful :D

Issue #632, opened 2021-04-30
[false negative] SCSS variables are partially broken


theKashey:
### Bug Description

This is not the issue of depcheck, but scss-parser - https://github.com/salesforce-ux/scss-parser/issues/38 - adding here for traceability

Behavior - scss-parser generates a run time exception parsing file content, as a result used packages are not tracked

### Code snippets

```scss
@use "~@ui/styles/mixin";
@use "~@ui/styles/media";

.button {
  @include mixin.until(media.$sm) { // fails here
    width: 100%;
  }
}
```

theKashey:
Still a problem

miluoshi:
This is not solved yet. Actually, there's no need for unmaintained `scss-parser`, because `sass` can parse scss files.

PR #710, opened 2022-04-19
feat: use current sass package to parse scss files

miluoshi:
There's no need for custom parser for `*.scss` files. `sass` package can handle `scss` syntax. This PR updates `sass` to the latest version and replaces deprecated `sass.renderSync` -> `sass.compile`.

Related: #632

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -7,7 +7,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest]
-        node-version: [10.x, 12.x, 14.x]
+        node-version: [14.x, 16.x, 18.x]
 
     runs-on: ${{ matrix.os }}
 
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ Depcheck not only recognizes the dependencies in JavaScript files, but also supp
 - [React JSX](http://facebook.github.io/react/docs/jsx-in-depth.html)
 - [CoffeeScript](http://coffeescript.org/)
 - [TypeScript](http://www.typescriptlang.org/) (with `typescript` dependency)
-- [SASS and SCSS](http://sass-lang.com/) (with `node-sass` dependency)
+- [SASS and SCSS](http://sass-lang.com/)
 - [Vue.js](https://vuejs.org/) (with `@vue/compiler-sfc` dependency)
 
 To get the syntax support by external dependency, please install the corresponding package explicitly. For example, for TypeScript user, install depcheck with `typescript` package:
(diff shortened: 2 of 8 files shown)


rumpl:
I <3 our dogfooding, it looks like the build is failing because `query-ast` is unused now :)

miluoshi:
Nice :) `query-ast` has been removed now as well.

rumpl:
Just kicked the CI, we now have some tests failing, could you take a look?

miluoshi:
I've updated the PR. `@import`/`@use` parsing was not done correctly in some cases. 

- I needed to update tested scss files for them to be valid (e.g. `@use` rules need to be always before `@import`).
- `sass` compiler is not reporting usage of its' internal modules (e.g. `@use "sass:math"`), so I've added regex matching against every file. It is not an optimal solution performance-wise. 
 
Imho a better solution to above performance problem would be that `sass` package would be reported as "used" by depcheck automatically if any `*.{scss,sass}` file is parsed, but then the following parts of spec file would have to be updated to not fail test:
```js
{
  name: 'support SASS/SCSS syntax',
  expected: {
    missing: {
      / ...
      sass: ['scss2.scss'], // <-- this should include each parsed sass file
    },
    using: {
      // ...
      sass: ['scss2.scss'], // <-- this should include each parsed sass file
    }
  },
```

miluoshi:
CI failed on one machine. Did it fail because of this error?

```
> cross-env NODE_ENV=test nyc mocha ./test ./test/special --timeout 20000 && nyc report --reporter=text-lcov > ./coverage/coverage.lcov

'cross-env' is not recognized as an internal or external command,
operable program or batch file.
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! depcheck@0.0.1 test-coverage: `cross-env NODE_ENV=test nyc mocha ./test ./test/special --timeout 20000 && nyc report --reporter=text-lcov > ./coverage/coverage.lcov`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the depcheck@0.0.1 test-coverage script.
```

I'm not sure what are problems with codecov, because I can't see the checks.

Anyway, I've also removed `node-sass` in the latest commit, because [it was not being used for some time](https://github.com/depcheck/depcheck/commit/04217610daf58b02509d75e04e589e1666b5fd6c)...

miluoshi:
Thank you. I'll check and fix the tests

rumpl:
Thanks! Don’t mind the codecov, it’s kinda broken and I think I’ll remove it at some point

miluoshi:
I can see it is failing only in Windows machine, but I have no idea why. I currently don't have access to a Windows machine to debug it. All the tests pass in Linux/MacOS.

I can get to Windows machine later this week to see what might be causing it.

PR #716, opened 2022-05-27
Fix typescript declaration for specials

dobesv:
They don't get the content as the first parameter like regular parsers do.  And they all seem to return a string array rather than an object.

### DIFF
--- a/index.d.tmpl
+++ b/index.d.tmpl
@@ -12,7 +12,7 @@ declare namespace depcheck {
   }
 
   type Parser = (content: string, filePath: string, deps: ReadonlyArray<string>, rootDir: string) => Node;
-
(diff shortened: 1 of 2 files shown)


index.d.tmpl
```diff
@@ -12,7 +12,7 @@ declare namespace depcheck {
   }
 
   type Parser = (content: string, filePath: string, deps: ReadonlyArray<string>, rootDir: string) => Node;
-
+  type SpecialParser = (filePath: string, deps: ReadonlyArray<string>, rootDir: string) => string[];
```
dobesv, line 15:
```suggestion
  type SpecialParser = (filePath: string, deps: ReadonlyArray<string>, rootDir: string) => Node | string[];
```

dobesv:
Actually this is incorrect - all the parsers are the same, none of them get file content as the first parameter.

PR #717, opened 2022-05-27
Handle scoped package bin with default name

dobesv:
When package.json specifies `"bin"` as just a string, it is treated as if they specified an object with the non-namespaced name of the package as the key and the bin value as the value.

See https://classic.yarnpkg.com/lang/en/docs/package-json/#toc-bin for docs

This one's been a bit of a pain to workaround because every time I run yarn it'll "optimize" the bin entry to use this feature, e.g. if I explicitly write something like this:

```
{
"name": "@foo/command",
"bin":  {
   "command": "dist/command.js"
}
```

After running `yarn` it rewrites it like this:

```
{
"name": "@foo/command",
"bin": "dist/command.js"
}
```

### DIFF
--- a/src/special/bin.js
+++ b/src/special/bin.js
@@ -17,7 +17,9 @@ function getBinaries(dep, dir) {
   const binMetadata = getCacheOrLoad(dep, dir);
 
   if (typeof binMetadata === 'string') {
-    return [[dep, binMetadata]];
+    // Use path.basename to discard any scope
+    // e.g. a package named "@foo/bar" creates a command named "bar"
+    return [[path.basename(dep), binMetadata]];
   }
 
   return lodash.toPairs(binMetadata);
--- a/test/special/bin.js
+++ b/test/special/bin.js
@@ -108,10 +108,22 @@ const testCases = [
     expected: ['binary-package'],
   },
   {
-    name: 'detect packages with single binary',
+    name: 'detect packages with single binary with default name',
     script: 'single-binary-package --argument',
     dependencies: ['single-binary-package'],
     expected: ['single-binary-package'],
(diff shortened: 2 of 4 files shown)


dobesv:
I pushed a fix which seems to have fixed the ubuntu builds, but not the windows ones.

rumpl:
Actually, I'm sorry but I don't understand the relation between your code and the PR description

src/special/bin.js
```diff
@@ -17,7 +17,7 @@ function getBinaries(dep, dir) {
   const binMetadata = getCacheOrLoad(dep, dir);
 
   if (typeof binMetadata === 'string') {
-    return [[dep, binMetadata]];
+    return [[path.basename(dep), binMetadata]];
```
dobesv, line 20: Desired transformation here:

- Unscoped package is left unchanged, e.g. `some-command`
- Scoped package discards scope, e.g. `@foo/some-command` -> `some-command`

dobesv:
I added a comment, let me know if that helps or not.

src/special/bin.js
rumpl: I see, but wouldn't that break things like `"bin": "bin/command-name.js"`?

dobesv: It would be nice to add a test but I'm not quite sure where and how best to do that.  I haven't found existing tests for this particular feature I could base a new test on.

dobesv:
I figured out how the tests work for this one and added a test.  If you think there's a case this will break let me know and I could add a test for that.

dobesv:
Hmm too bad it would have been nice to get this merged in

PR #718, opened 2022-06-01
Tweak Parser typescript definition

dobesv:
- Allow string array return
- Use `Record` instead of mapped type.
- Remove `content` parameter - this doesn't seem to ever be passed in

### DIFF
--- a/index.d.tmpl
+++ b/index.d.tmpl
@@ -9,7 +9,7 @@ declare function depcheck<T>(
 declare namespace depcheck {
   type Node = Record<string, any>;
 
-  type Parser = (content: string, filePath: string, deps: ReadonlyArray<string>, rootDir: string) => Node | string[];
+  type Parser = (filePath: string, deps: ReadonlyArray<string>, rootDir: string) => Node | string[];
 
   type Detector = (node: Node) => ReadonlyArray<string> | string;
 
--- a/index.d.tmpl
+++ b/index.d.tmpl
@@ -7,11 +7,9 @@ declare function depcheck<T>(
 ): Promise<T>;
 
 declare namespace depcheck {
-  interface Node {
-    [key: string]: any;
-  }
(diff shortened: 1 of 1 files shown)


index.d.tmpl
```diff
@@ -7,11 +7,9 @@ declare function depcheck<T>(
 ): Promise<T>;
 
 declare namespace depcheck {
-  interface Node {
-    [key: string]: any;
-  }
+  type Node = Record<string, any>;
 
-  type Parser = (content: string, filePath: string, deps: ReadonlyArray<string>, rootDir: string) => Node;
+  type Parser = (content: string, filePath: string, deps: ReadonlyArray<string>, rootDir: string) => Node | string[];
```
dobesv, line 12:
```suggestion
  type Parser = (filePath: string, deps: ReadonlyArray<string>, rootDir: string) => Node | string[];
```

index.d.tmpl
```diff
@@ -7,11 +7,9 @@ declare function depcheck<T>(
 ): Promise<T>;
 
 declare namespace depcheck {
-  interface Node {
-    [key: string]: any;
-  }
+  type Node = Record<string, any>;
 
-  type Parser = (content: string, filePath: string, deps: ReadonlyArray<string>, rootDir: string) => Node;
+  type Parser = (filePath: string, deps: ReadonlyArray<string>, rootDir: string) => Node | string[];
```
rumpl, line 12: Not sure I understand this change, the parser doesn't have the `content` parameter? Seems weird to change the type definition and not the code at the same time

rumpl: Oh! Ok 😅

jtbandes:
Hi, is there any plan to publish this in a release?

rumpl:
Hi @jtbandes I'll try and push one this week 👍

Issue #723, opened 2022-06-03
Support babel config that exports a function


dobesv:
### Feature request description

When you define a `babel.config.js` you can return a function, e.g.

https://gist.github.com/gtzilla/a6ae9777a987df1ee1357df51bf9cd61

When parsing this config file, if it a returns a function the function has to be called and provided a stub implementation of that babel api.

See https://babeljs.io/docs/en/config-files#config-function-api

PR #724, opened 2022-06-03
If babel config is a function, call it to get config

dobesv:
When you define a `babel.config.js` you can return a function, e.g.

https://gist.github.com/gtzilla/a6ae9777a987df1ee1357df51bf9cd61

When parsing this config file, if it a returns a function the function has to be called and provided a stub implementation of that babel api.

See https://babeljs.io/docs/en/config-files#config-function-api

Fix for https://github.com/depcheck/depcheck/issues/723


rumpl:
Code looks good but it would be great to have a test for this

dobesv:
I added a test case now

rumpl (approved):
Nice

Issue #720, opened 2022-06-01
Support typescript "extends" and "types" fields


dobesv:
### Feature request description

In TypeScript you can referenced your dependencies in the `extends` field, for example:

```
"extends": "@some/package/tsconfig.base.json"
```

or just

```
"extends": "@some/package"
```

So when processing typescript files we need to look into the `tsconfig.json` file and if we see an `extends` key we need to treat that like an import.

Also, in `tsconfig.json` you can specify a `types` field which sort of implicitly imports those types into every file, e.g.

```
"types": ["node", "jest"]
```

Creates dependencies on `@types/node` and `@types/jest`

Example implementation:

```
const requirePackageName = require('require-package-name');

const tsconfigParser: Parser = (filePath, deps, rootDir) => {
  const content = readFileSync(filePath, {encoding: 'utf8'});
  const foundDeps: string[] = [];
  const tsconfigJson = JSON.parse(content);
  const types = tsconfigJson.compilerOptions?.types;
  if (types) {
    for (const pkg of types) {
      foundDeps.push(`@types/${pkg}`);
    }
  }
  if (tsconfigJson.extends) {
    foundDeps.push(tsconfigJson.extends);
  }
  return foundDeps.map(p => requirePackageName(p)).filter(Boolean);
};

```

dobesv:
not stale

BenJackGill:
+1 for this feature.

I have a monorepo made with `turborepo` and it uses the "extends" pattern a lot.

This would fix this false positive.

PR #725, opened 2022-06-03
Add support for some tsconfig.json fields

dobesv:
Detect dependencies for the "types" and "extends" fields in tsconfig.json

Fixes https://github.com/depcheck/depcheck/issues/720

### DIFF
--- a/src/component.json
+++ b/src/component.json
@@ -7,6 +7,7 @@
     "graphql",
     "jsx",
     "sass",
+    "tsconfig",
     "typescript",
     "vue",
     "svelte"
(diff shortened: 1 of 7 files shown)


dobesv:
not stale

Nokel81:
Would we be able to get a release with this fix?

OnkelTem:
So this one was not yet released?
I'm on 1.4.5, but **depcheck** keeps reporting that:

```
Unused devDependencies
* @tsconfig/node16
```

despite I have this in my **tsconfig.json**:

```json
{
  "extends": "@tsconfig/node16/tsconfig.json"
}
```

PR #731, opened 2022-06-08
Allow jest import errors to escape and be reported as parse errors

dobesv:
I spent quite some time trying to figure out why some jest dependencies were not being picked up, and it turns out it was because of an error loading the jest config.  Since depcheck does handle and report errors from the parser nicely it seems better to let these import errors pass out to depcheck to be handled that way rather than silently suppressing them.


dobesv:
Hmm it seems like there's some tests and code indicating that suppressing these errors was specifically desired at some point.  What do you think - is it better to suppress errors when loading js files or to allow them to pass out and be reported?

PR #732, opened 2022-06-08
Use createRequire to import files

dobesv:
Using `createRequire` from `module` allows module resolution to proceed as if your code is running inside that project.  This means that features like `require.resolve` or `require` will be resolved relative to that project using its resolution rules.  This can avoid errors in cases where depcheck is not running as a file inside that project, or in a Yarn PnP environment where modules are not allowed to `require` other modules without a declared dependency.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -16,11 +16,16 @@ import {
 
 function registerTs(rootDir) {
   if (!require.extensions['.ts']) {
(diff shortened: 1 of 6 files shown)


dobesv:
It looks like `createRequire` requires node 12+.  Should I make some version of this that falls back to the normal `require` for node 10?  Or would you consider dropping support for node v10 as it's quite old at this point?

PR #745, opened 2022-08-05
convert object to array for babel support

Robbilie:
babel requires arrays, other parsers have this corrected, the typescript one doesnt and latest babel versions use destructuring which fails the whole parser rather than ignore the plugin

### DIFF
--- a/src/parser/typescript.js
+++ b/src/parser/typescript.js
@@ -16,7 +16,7 @@ export default async function parseTypescript(filename) {
       'classProperties',
       'classPrivateProperties',
       'classPrivateMethods',
-      // { decorators: { decoratorsBeforeExport: true } },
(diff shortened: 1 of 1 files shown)


Robbilie:
previously
```
if (Array.isArray(plugin)) {
      return plugin[0] === name;
    } else {
      return plugin === name;
    }
```
    
more recent version
```
    if (typeof p === "string") {
      return expectedOptionsIsEmpty && p === expectedName;
    } else {
      const [pluginName, pluginOptions] = p;
```
      
which crashed with p not being enumerable…
if possible a fix for 0.8.x and 0.9.x would be nice…

am-a:
Any chance this can be looked at again? As this effectively prevents upgrading to TS 4.9+

PR #749, opened 2022-08-19
Adds Vue3 setup support

devan1011:
Adds support for Vue 3 composition `<script setup>` components. As `<script>` and `<script setup>` are allowed in the same file I just concat the two files together as it's only import parsing that is required.

Also adds supports for typescript parsing in Vue components.


xibman:
👍  for this one !

BenJackGill:
Would also love Vue 3 support.

+1 from me also!

devan1011:
Main difference is I concat the script and setupScript as in Vue 3 it is valid to have for example:
```vue
<script lang="ts">
export default {
  name: 'ComponentName',
};
</script>

<script setup lang="ts">
// ...
</script>
```
I had a use case where there where imports in both scripts, and didn't want to loose the deps in the first script tag.

rumpl (approved):

PR #760, opened 2022-11-23
add esm support for bin packages

dword-design:
The PR changes `loadModuleData` so that it supports ESM packages that do not export `package.json`.


dword-design:
@rumpl Done. You can squash the commits when merging, right?

rumpl:
I can and I did, but I had to activate the feature... Thanks!

dword-design:
@rumpl Can we publish a new version to npm so this fix is in? 😀

PR #788, opened 2023-03-31
Upgrade Babel parser to support Typescript 5.0

### LINKED ISSUE #768: Breaks when using `satisfies` operator in TypeScript
I think this is due to the used babel version, it has been released in `7.20.0` https://github.com/babel/babel/pull/14211

Karfroth:
This PR upgrades Babel parser to 7.21.3, which supports Typescript 5.0. As it is minor version upgrade, I would expect that there will be no breaking changes.

This will help checking the typescript files with new keywords, such as `satisfies`. Hence, this PR can fix #768

To ensure the new syntex support, this PR also adds a simple function to the index.ts file in the test.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
   "license": "MIT",
   "readmeFilename": "README.md",
   "dependencies": {
-    "@babel/parser": "7.16.4",
+    "@babel/parser": "7.21.3",
     "@babel/traverse": "^7.12.5",
     "@vue/compiler-sfc": "^3.0.5",
     "callsite": "^1.0.0",
--- a/package.json
+++ b/package.json
@@ -49,8 +49,8 @@
   "license": "MIT",
   "readmeFilename": "README.md",
   "dependencies": {
-    "@babel/parser": "7.21.3",
-    "@babel/traverse": "^7.12.5",
+    "@babel/parser": "7.22.5",
+    "@babel/traverse": "7.22.5",
     "@vue/compiler-sfc": "^3.0.5",
     "callsite": "^1.0.0",
     "camelcase": "^6.2.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,8 +9,8 @@
       "version": "0.0.1",
       "license": "MIT",
       "dependencies": {
-        "@babel/parser": "7.21.3",
-        "@babel/traverse": "^7.12.5",
+        "@babel/parser": "7.22.5",
+        "@babel/traverse": "7.22.5",
         "@vue/compiler-sfc": "^3.0.5",
         "callsite": "^1.0.0",
         "camelcase": "^6.2.0",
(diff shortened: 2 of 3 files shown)


danyf90 (approved):

jportner:
I tried using this upgraded dependency (via a yarn dependency resolution) and it seems to have completely broken TypeScript parsing on my project. I get a bunch of "TypeError: p is not iterable" errors as a result.

It seems this was a previously documented issue (https://github.com/depcheck/depcheck/issues/688) and it was [fixed in depcheck 1.4.3](https://github.com/depcheck/depcheck/compare/1.4.2...1.4.3) by pinning @babel/parser to exactly 7.6.14. There's a related discussion here too (https://github.com/depcheck/depcheck/issues/687).

TL;DR unfortunately it seems that upgrading @babel/parser to 7.21.3 re-introduces this problem :(

Karfroth:
Thanks for the comment. I will try check the root cause.

Karfroth:
@jportner I tried to dig more and I believe this PR will not cause the same issue again. because the fix for the root cause is already merged to the main branch with #745

jportner:
> @jportner I tried to dig more and I believe this PR will not cause the same issue again. because the fix for the root cause is already merged to the main branch with #745

Oh I see, that has been merged but not released yet.
I was wondering why your unit test passed!

Thanks for double-checking, and apologies for the false alarm 😄

danyf90:
@rumpl do you think we can merge this and get a new release to fix support for TS5? 🙏

danyf90:
@rumpl / @jportner  do you think we can merge this and get a new release to fix support for TS5? 🙏

seansfkelley:
As a workaround, if you're using Yarn, you can both peg the resolved version of the Babel parser (a la https://github.com/depcheck/depcheck/pull/788#issuecomment-1505676628) and then `yarn patch depcheck` to manually patch in the fix to #688, which is thankfully very easy to apply in the transpiled `.js`.

package.json
```diff
@@ -49,7 +49,7 @@
   "license": "MIT",
   "readmeFilename": "README.md",
   "dependencies": {
-    "@babel/parser": "7.16.4",
+    "@babel/parser": "7.21.3",
```
znarf, line 52:
```suggestion
    "@babel/parser": "7.22.5",
```

Maybe update to the latest.

package.json
```diff
@@ -49,7 +49,7 @@
...
     "@babel/traverse": "^7.12.5",
```
znarf, line 53:
```suggestion
    "@babel/traverse": "7.22.5",
```

Maybe update that one too for consistency.

package.json
Karfroth: Thank you. I did this manually to update lock file as well.

package.json
Karfroth: Same as `@babel/parser`, I did this manually to update lock file.

znarf (approved):
Good for me.

danyf90:
@znarf @jportner is there any blocker for merging and releasing this? 🙏

rumpl:
> @znarf @jportner is there any blocker for merging and releasing this? 🙏

None, I’ll do it this weekend when I get back from my PTO

danyf90:
Thanks for merging this @rumpl 🙏 
Do you think we can do a release containing this change? 🙏

jtbandes:
I'm also waiting for this change to be released 🎉  thank you!

PeKne:
Thank you for fixing this. Looking forward to the release 🙏.

pkishorez:
@rumpl thanks a lot for your efforts.
I could see some changes merged to main, yet no release created since last year!
Is there any issue regarding the same?

PR #790, opened 2023-04-19
Add stylelint as a new Special

Josehower:
This PR intends to add stylelint as a special very similar to eslint special. It's still a work in progress and I am open to anyone that wants to collaborate in finishing it.


AdamPierwszy:
Oooooooooo

PR #793, opened 2023-04-27
feat: support tslib as special

### LINKED ISSUE #771: tslib reported
### Bug Description

we use https://www.npmjs.com/package/tslib with this config in the tsconfig.base.json

The tslib is reported in the unused dependencies.

### Code snippets

tsconfig.base.json
```
  "importHelpers": true,
```

Where the dependency is listed in package.json:

```
  "tslib": "2.2.0",
```

### Versions

- `node -v`: v18.12.1
- `npm -v`: 8.19.2
- `depcheck --version`:1.4.3

bangbang93:
fix #771

### DIFF
--- a/src/special/tslib.js
+++ b/src/special/tslib.js
@@ -1,0 +1,16 @@
+import { basename } from 'path';
+import { getContent } from '../utils/file';
+
+const tsconfigPattern = /tsconfig(?:\.[^.]+)*\.json/;
+export default async function parseTslib(filename) {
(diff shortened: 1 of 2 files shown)


znarf:
Why not implement this in the [tsconfig parser](https://github.com/depcheck/depcheck/blob/main/src/parser/tsconfig.js) instead of introducing a new special?

bangbang93:
here is reimplement in tsconfig parser

PR #798, opened 2023-05-14
chore: downgrade sass to unbreak node 10

legobeat:
#710 upgraded version constraint of dependency `sass` to `^1.50.1`. However, while `depcheck` still indicates support for nodejs 10, the latest version of `sass` with support for node 10 is `1.49.0`. 

This draws down the range so current main branch (and maintenance fixes) can be released straight from `main` to `1.x` before doing a semver-major.

Also includes applies of `npm audit --fix` and `npm dedupe` on transient deps.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -71,27 +71,32 @@
       }
     },
     "node_modules/@babel/cli": {
-      "version": "7.12.1",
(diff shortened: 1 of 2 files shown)


znarf:
Node 10 was "end of life" on 2021-04-30. I don't think it should be supported.

rumpl:
We no longer depend on sass, let me close this one

Issue #779, opened 2023-02-14
A dynamic import with template string literal is not found by depcheck


ArtskydJ:
### Bug Description

Depcheck is not aware that a dependency is used if it's imported with a dynamic import with a template string literal.

### Code snippets

Where the dependency is used:

```js
const micromark = await import(`micromark`)
```

Where the dependency is listed in package.json:

```
{
	"dependencies": {
		"micromark": "^3.1.0",
		...
	}
}
```

### Versions

- `node -v`: 16.19.0
- `npm -v`: 8.19.3
- `depcheck --version`: 1.4.3

### Extra info

The exact same code, but replacing ticks with quotes works as expected.

ArtskydJ:
With #810 merged, this will be fixed upon next npm release.  Current release is 1.4.3.

ArtskydJ:
Fixed in v1.4.4 according to the [changelog](https://github.com/depcheck/depcheck/releases/tag/v1.4.4).  I confirmed it's working in v1.4.6.

rumpl:
🎉

PR #810, opened 2023-06-20
Fix dynamic import with a template literal #779

ArtskydJ:
Depcheck doesn't catch this

```js
import(`whatever`)
```

But with this PR, it now will catch that import.


znarf (approved):
Looks good at first glance.

rumpl (approved):

Issue #814, opened 2023-06-29
[Bug] Webpack: False alerts with TypeScript >= 4.9


vivi90:
## **False alerts**

Following webpack loaders are false alerts (btw. all loaders, i use):

- babel-loader
- css-loader
- file-loader
- html-loader
- raw-loader
- source-map-loader
- style-loader

And the following polyfills:

- buffer
- es6-promise
- core-js

## **Additional false alerts** due to this bug

Since depcheck is not able to parse my `webpack.config.ts` due to the bug, i get the following additional false alerts:

- `copy-webpack-plugin`
- `fork-ts-checker-webpack-plugin`
- `html-webpack-plugin`
- `mustache`
- `@types/mustache`
- `office-addin-dev-certs`
- `tsconfig-paths-webpack-plugin`
- `@types/webpack`
- `webpack-dev-server`
- `@types/node`
- `crypto-browserify`
- `stream-browserify`

## The bug

TypeScript 4.9 introduced the `satisfies` operator, using it in my `webpack.config.ts`:

```ts
import CopyWebpackPlugin from "copy-webpack-plugin";
import ForkTsCheckerWebpackPlugin from "fork-ts-checker-webpack-plugin";
import HtmlWebpackPlugin from "html-webpack-plugin";
import { render as mustacheRender } from "mustache";
import { getHttpsServerOptions } from "office-addin-dev-certs";
import TsconfigPathsPlugin from "tsconfig-paths-webpack-plugin";
import { Configuration as WebpackConfiguration, DefinePlugin, ProvidePlugin } from "webpack";
import { Configuration as WebpackDevServerConfiguration, ServerOptions } from "webpack-dev-server";

interface Configuration extends WebpackConfiguration {
  devServer?: WebpackDevServerConfiguration;
}

...

type EnvironmentSettings = {
  [key in EnvironmentName]: SomeSettings;
};

enum EnvironmentName {
  EXAMPLE1 = "environment1",
  EXAMPLE2 = "environment2",
  ...
}

type AvailableEnvironment = keyof typeof payload;

const payload = {
  environment1: {
    ...
  },
  environment1: {
    ...
  },
} satisfies Partial<EnvironmentSettings>;

...

module.exports = async (env: typeof process.env, options: Configuration["devServer"]) => {
  ...

  const config: Configuration = {
    devtool: "source-map",
    entry: {
      polyfill: ["core-js/stable", ...],
      ...
    },
    output: {
      ...
      clean: true,
    },
[...]

PR #815, opened 2023-06-29
Update typescript to 5.3.3

znarf:
May fix https://github.com/depcheck/depcheck/issues/814

### DIFF
--- a/package.json
+++ b/package.json
@@ -100,7 +100,7 @@
     "proxyquire": "^2.1.3",
     "should": "^13.2.3",
     "svelte": "^3.38.2",
-    "typescript": "^4.0.5"
+    "typescript": "^5.1.6"
   },
   "nyc": {
     "sourceMap": false,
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -101,7 +101,7 @@
     "proxyquire": "^2.1.3",
     "should": "^13.2.3",
     "svelte": "^3.38.2",
-    "typescript": "^4.0.5"
+    "typescript": "^5.1.6"
```
vivi90, line 104: Upgrade to TypeScript 5 is not neccessary and may break some type definitions.
Please use `^4.9.0`.

vivi90 (changes requested):
Please change to `^4.9.0` or `^4.9.5` instead.
But better to `^4.9.0` since that is the required version.

rumpl:
> But better to `^4.9.0` since that is the required version.

Required for what?

>  TypeScript 5 may break some type definitions.

Can you be more specific please?

znarf:
I think the tool should generally try to update its dependencies continuously to their latest version.

Here, if there is no test broken, we can assume that there is no breaking change and can move forward.

If we want to document a particular behavior and prevent against breaking changes, we should ideally have that covered with a test.

rumpl (approved):

PR #826, opened 2023-08-22
feat(tsconfig): support tsconfig.json files with comments

miluoshi:
Built-in JSON.parse() doesn't allow json with comments, but `tsconfig.json` files support comments. By using `json5` instead of native JSON we prevent errors when trying to parse such json files with comments

### DIFF
--- a/src/parser/tsconfig.js
+++ b/src/parser/tsconfig.js
@@ -1,10 +1,12 @@
 const requirePackageName = require('require-package-name');
 const { readFileSync } = require('fs');
+const JSON5 = require('json5');
 
 export default function tsconfigParser(filePath, deps) {
   const content = readFileSync(filePath, { encoding: 'utf8' });
   const foundDeps = [];
-  const tsconfigJson = JSON.parse(content);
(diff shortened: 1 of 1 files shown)


miluoshi:
It would probably be a good idea to add a test for this as well.

rumpl:
What does tsc use to parse its tsconfig.json file?

miluoshi:
> What does tsc use to parse its tsconfig.json file? It might be a good idea to use the same thing if possible

tsc uses `jsonc-parser` [here](https://github.com/microsoft/TypeScript/blob/5439c8111de50fa6ea239f95fc23c76dc868fc07/scripts/build/utils.mjs#L78). But since json5 (which this project is already using) is [more widely used](https://npmtrends.com/json5-vs-jsonc-parser) package and `jsonc-parser` supported syntax is a subset of `json5` supported syntax, it made sense to me to just reuse `json5`.

rumpl:
Fair enough, could you just add a comment in this file saying what you just said, kinda explaining why we are using a different lib than TS. For our future selves and people coming here are reading the code :)

znarf:
Duplicate with https://github.com/depcheck/depcheck/pull/806

rumpl (approved):

PR #829, opened 2023-08-24
feat: add --quiet flag to suppress No depcheck issue messages

### LINKED ISSUE #759: Flag to make logs readable and less verbose
Description:
Reading the logs generated by dep check is very time-consuming, I am personally not interested in seeing hundreds of lines of  `YN0000: [@package]: No depcheck issue` to find an unused dependency.
Removing the verbose flag doesn't help either since it wouldn't display the dependency that actually has an issue and will still log hundreds of lines of `No depcheck issue`.

it would be more useful to remove the log for  `No depcheck issue` all together  when removing the `--verbose` flag.

My project is huge and this would really help.

openam:
Closes #759

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,8 @@ All of the arguments are optional:
 
 `--ignore-patterns`: Comma separated patterns describing files to ignore. Patterns must match the .gitignore [spec](http://git-scm.com/docs/gitignore). Example, `--ignore-patterns=build/Release,dist,coverage,*.log`.
 
+`--quiet`: Suppress the "No depcheck issue" log. Useful in a monorepo with multiple packages to focus only on packages with issues.
+
 `--help`: Show the help message.
 
 `--parsers`, `--detectors` and `--specials`: These arguments are for advanced usage. They provide an easy way to customize the file parser and dependency detection. Check [the pluggable design document](https://github.com/depcheck/depcheck/blob/master/doc/pluggable-design.md) for more information.
--- a/src/cli.js
+++ b/src/cli.js
@@ -86,7 +86,7 @@ function print(result, log, opt, rootDir) {
         lodash.isError(value) ? value.stack : value,
       ),
     );
-  } else if (noIssue(result)) {
+  } else if (noIssue(result) && !opt.quiet) {
     log('No depcheck issue');
   } else {
     const deps = prettify(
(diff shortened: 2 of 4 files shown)


rumpl:
In my mind a `--quiet` would make depcheck not print anything at all, could we update it so that `--quiet` doesn't print anything?

openam:
@rumpl That sounds more like a `--silent` tag. The point of this PR was to make it so that people using a monorepo wouldn't have to look through 100 lines of `No depcheck issue` noise to find the one line that has an error. e.g.

```
many more
...
constants: No depcheck issue
permissions: No depcheck issue
utils: No depcheck issue
errors: No depcheck issue
logger: No depcheck issue
aws-sqs: No depcheck issue
aws-s3: No depcheck issue
koa-logger: No depcheck issue
opsgenie: No depcheck issue
config: Missing dependencies
config: * logger: ./index.js
geo: No depcheck issue
proxy: No depcheck issue
ably: No depcheck issue
dashboard: No depcheck issue
authenticator: No depcheck issue
...
plus many more
...
lerna ERR! Received non-zero exit code 255 during execution
lerna success exec Executed command in 108 packages: "depcheck"
```

vs

```
config: Missing dependencies
config: * logger: ./index.js
lerna ERR! Received non-zero exit code 255 during execution
lerna success exec Executed command in 108 packages: "depcheck --quiet"
```

seansfkelley:
Throwing my hat in the ring here as a commenter on the original issue: I agree on the distinction between `--quiet` and `--silent` (though can't really imagine when I would want to use the latter with a tool like depcheck). When I give `--quiet` to some arbitrary tool, I typically expect only "essential" output to be printed (or: output requiring an action from me). Maybe there is more non-essential logging that should be covered by `--quiet` than just this line, but I definitely _do_ want to see in which file the issues were found.

rumpl (approved):

openam:
@rumpl Is there anything I need to do to get a version released that contains this change? Unfortunately this was merged in like 3 days after the last release.

rumpl:
You only need to ask, I'll make a new release tomorrow ;)

openam:
@rumpl would you please make a release with this `--quiet` flag.

rumpl:
@openam Version 1.4.6 with this option was just released!

PR #835, opened 2023-09-12
--quiet: don't print an empty line if nothing to report

seansfkelley:
Without this change, setting --quiet would hit the third branch of the reporting code, and report... nothing, with an empty line.

Follow-up to https://github.com/depcheck/depcheck/pull/829.


rumpl (approved):
Damn! Thanks

PR #836, opened 2023-09-13
Support package.json subpath imports

dobesv:
If you specify an "imports" field in package.json it is used to resolved imports starting with "#" in the code.

For more details, see https://nodejs.org/api/packages.html#subpath-imports


test/fake_modules/package_import_map/index.js
```diff
@@ -0,0 +1,20 @@
+/* eslint-disable no-unused-vars */
+
+// This file includes some experimental ES7 syntax enabled in Babel by default. Reference: https://babeljs.io/docs/usage/experimental/
```
rumpl, line 3: Is this true? Or a copy/paste?

test/fake_modules/package_import_map/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "imports":{
+    "#simple": "simple-resolved",
+    "#subpath": "subpath-resolved/subpath",
+    "#conditional": {
+      "node": "conditional-node",
+      "types": "@types/conditional"
```
rumpl, line 7: I don't see `"types"` in the list of conditions that are handled https://nodejs.org/api/packages.html#conditional-exports

test/fake_modules/package_import_map/package.json
```diff
@@ -0,0 +1,31 @@
...
+    },
+    "#multi-conditional": {
+      "node": {
+        "test": "conditional-node-test"
```
rumpl, line 11: Same question, what is the `"test"` condition?

test/fake_modules/package_import_map/index.js
dobesv: Ah definitely copy/paste, sorry

rumpl: No problem, I wasn't sure, haven't looked at javascript language changes in years :D

test/fake_modules/package_import_map/package.json
dobesv: That's from TypeScript, e.g. https://www.typescriptlang.org/docs/handbook/esm-node.html

test/fake_modules/package_import_map/package.json
dobesv: In our project we add some custom conditions by using `enhanced-resolve` where we can specify whatever condition names we want. So in tests, we have `test` condition set, in storybook we set `storybook`.  This allows some imports to be redirected according to the environment.

test/fake_modules/package_import_map/package.json
rumpl: Oooh, gotcha, thanks

test/fake_modules/package_import_map/index.js
dobesv: I pushed a cleaned up version of this one

test/fake_modules/package_import_map/package.json
rumpl: I see, and node/other tools just ignore conditions they don't know about right?

dobesv: Yeah basically any tool treats the conditions as a set and if it finds an object it descends into the first key that is in that set.  If it finds a string, it restarts resolution replacing the first path component with that string and maybe doing wildcard replacement if any.  If it hits a dead end, the import fails.

dobesv:
I improved the code docs for that import map thing for the benefit of future maintainers trying to figure out what it's all about.

src/check.js
```diff
@@ -40,14 +40,61 @@ function detect(detectors, node, deps) {
...
+        resolvedDeps.push(requirePackageName(v));
+      } else if (typeof v === 'object') {
+        Object.values(v).forEach(accumulateDeps);
+      }
+    }
+  }
+
+  if (dep.startsWith('#')) {
+    Object.entries(importsMap).forEach((m) => {
+      if (`${dep}/`.startsWith(`${m[0]}/`)) {
```
rumpl, line 71: Can't this be simplified? 

```suggestion
      if (dep.startsWith(m[0])) {
```

dobesv: I'm trying to avoid a case where you have, say:

```
"imports": {
  "#foobar": "some-other-package"
  "#foo": "some-package",
}
```

If you import `#foo` or `#foo/whatever` it should only match a key of `#foo` with no extra characters after *except* a `/`.

dobesv: Maybe I wrote it in a confusing way, though ... maybe I need to split on slash and compare explicitly 🤔

dobesv: I rewrote it to be a lot more clear what is going on.

rumpl: Ah, right, but then even adding a `/` won't do us no good. You can put 

```
  "imports": {
    "#toto": "./file1.js",
    "#toto/hm": "./file2.js"
  }
```
in your package.json file and in this case depcheck could be wrong no?

Do we even need a `startsWith`? Shouldn't the imports in the code be exactly the same as defined in package.json?

rumpl: Right so I just checked something, I hope I'm checking the right thing, it's getting late here :D 

With `"#ho": "./ha"` in my imports in package.json, and `ha` being a directory I can't `require('#ho/pf.js'), I get an error:

```
TypeError [ERR_PACKAGE_IMPORT_NOT_DEFINED]: Package import specifier "#ho/pf.js" is not defined in package /private/tmp/ntest/package.json imported from /private/tmp/ntest/index.js
```

rumpl: Also, you could have 

```
"#ho": "./foo.js"
"#ho/pf": "./ha/pf.js"
```
in your imports.

So I'm beginning to think this should really be a `===` check.

dobesv: Ah, true, actually to import a subpath generically it would be more like:

```
"imports": {
"#ho/*": "ha/*"
}
```

So to handle this exactly to spec I'd have to do the wildcard processing.  Hmm.

rumpl: Ha, indeed wildcard works! Is there a real spec somewhere? I didn't see any of that in the PR description. If it's a regex then why don't we check it as a regex?

If we `dep.match(m)` then there is no more `#foo/thing` and `#foobar` shenanigans.

rumpl: Right, so it seems only `*` is permitted? https://nodejs.org/api/packages.html#subpath-patterns

rumpl: For future people reading this, here is what node is doing

https://github.com/nodejs/node/blob/4e0184294362b477e52c996d304e6fe0d6824964/lib/internal/modules/esm/resolve.js#L612-L670

dobesv: I updated it to handle wildcards the same way as documented in the node docs (I hope!).

dobesv: Or at least, hopefully it's close enough to avoid most false reports of unused deps that are actually used indirectly via this feature.

rumpl: Yeah, I'm OK with the latest code you pushed, it's might not be 100% in line with node but at least it fixes a whole class of errors we had. If people stumble on a difference between the implementations they will (hopefully) open issues :)

Thanks for working on this 👍

rumpl (approved):
LGTM, @dobesv if you don't mind, could you squash your commits before I merge?

dobesv:
I'm AFK now, I'll have to do it tomorrow.  Doesn't GitHub have a "squash and merge" option?  Or does that require a paid plan?

dobesv:
I squashed the commits

rumpl:
Thanks!

PR #840, opened 2023-10-12
Memoize `loadModuleData` to speed up depcheck runs

pastelsky:
While running depcheck on a large project with 100s of packages, I found that depcheck spends a lot of time resolving the package.jsons for path token it finds.

Screenshot 2023-10-12 at 6 16 07 PM

Memoizing the function cut down time from ~ 59s to ~37s

### DIFF
--- a/src/special/bin.js
+++ b/src/special/bin.js
@@ -2,19 +2,13 @@ import path from 'path';
 import lodash from 'lodash';
 import { loadModuleData, getScripts } from '../utils';
 
-const binaryCache = {};
-
-function getCacheOrLoad(dep, dir) {
-  const index = `${dir}/${dep}`;
-  if (!binaryCache[index]) {
-    const metadata = loadModuleData(dep, dir).metadata || {};
-    binaryCache[index] = metadata.bin || {};
-  }
-  return binaryCache[index];
+function getBinMetadata(dep, dir) {
+  const metadata = loadModuleData(dep, dir).metadata || {};
+  return metadata.bin || {};
 }
 
 function getBinaries(dep, dir) {
-  const binMetadata = getCacheOrLoad(dep, dir);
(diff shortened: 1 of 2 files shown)


rumpl:
This is nice! I guess we don't need this cache any more then? https://github.com/depcheck/depcheck/blob/b83d38b9d5589403de2ee14b589126818fd2c13f/src/special/bin.js#L7

rumpl:
Could you show the flamegraph before and after this change?

pastelsky:
#### Before
Screenshot 2023-10-12 at 10 00 27 PM


#### After
Screenshot 2023-10-12 at 10 00 20 PM

pastelsky:
> This is nice! I guess we don't need this cache any more then?
> 
> https://github.com/depcheck/depcheck/blob/b83d38b9d5589403de2ee14b589126818fd2c13f/src/special/bin.js#L7

I'm not exactly sure what this cache was meant for? FWIW, I'm invoking depcheck through its Node API so its possible its not hitting that.

rumpl:
> I'm not exactly sure what this cache was meant for? FWIW, I'm invoking depcheck through its Node API so its possible its not hitting that.

That cache is used when you invoke the `depcheck` cli, with the metadata loading memoized we don't need that cache any more

rumpl:
If you don’t mind removing it that would be great 👍

if you could also squash your commits please I would appreciate it

thanks

pastelsky:
Done. 

There's still an opportunity to optimize further — module resolution remains the single largest time-taking step in what `depcheck` does — so it might be worth profiling further, but for now, the cache avoids a very large part of the cost.

pastelsky:
@rumpl let me know if this looks good to go!

rumpl (approved):

pastelsky:
Thanks @rumpl. Do you mind releasing these changes?

PR #847, opened 2023-11-08
fix: detect correctly package name when importing nested file in @types package

gmeligio:
## Motivation
Currently `depcheck` will detect if a type is being imported and if the corresponding `@types/package-x` exists.

Some `@types` packages have types in more files other than `index.d.ts`. For example, `mdx-js` stores types in a file `mdx/types` as showed in https://mdxjs.com/docs/getting-started/#types . An example snippet that works for the TypeScript compiler but is reported as missing in `depcheck`.
```typescript
import type {MDXComponents} from 'mdx/types.js'
```

The TypeScript compiler correctly detects in this case that the package `@types/mdx` can be resolved.

While `depcheck` reports
```
Missing dependencies
* mdx: ./src/app/mdx/components/index.tsx
```

## Changes

This PR adds support for importing type declarations from nested files in `@ŧypes` packages like 

```typescript
import type { Bar } from 'another-typeless-module/nested-declaration-file';
```

Generally, it will extract the first path segment from the package path (`another-typeless-module` in the example) and use it for resolution checking. This works because the [naming rules](https://github.com/DefinitelyTyped/DefinitelyTyped?tab=readme-ov-file#naming) don't allow the substring after `@types` to contain slashes.

## Test

Added a new test to cover the scenario proposed.

1. `npm run lint` didn't introduce any new problems (there is an existing one in `src/utils/index.js#L27:10)
1. `npm run test` passed all tests
1. `npm run compile && npm run component && npm run depcheck && npm run depcheck-json` worked OK.

### DIFF
--- a/src/detector/importDeclaration.js
+++ b/src/detector/importDeclaration.js
@@ -5,13 +5,14 @@ export default function detectImportDeclaration(node, deps) {
     return [];
   }
 
-  // TypeScript "import type X from 'foo'" - doesn't need to depend on the
+  // TypeScript "import type X from 'foo'" and "import type X from 'foo/bar'"- doesn't need to depend on the
   // actual module, instead it can rely on `@types/<module>` instead.
-  if (
-    node.importKind === 'type' &&
-    deps.includes(`@types/${node.source.value}`)
-  ) {
-    return [`@types/${node.source.value}`];
+
+  const packageName = node.source.value.split('/')[0];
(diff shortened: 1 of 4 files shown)


rumpl (approved):
Awesome, thanks

gmeligio:
Thank you @rumpl for the prompt review!

rumpl:
Thank you @gmeligio for a nice and easy PR :) 

I'll release this tomorrow

PR #807, opened 2023-06-08
When parsing typescript, only load babel jsx plugin when the file is .tsx

znarf:

PR #849, opened 2023-11-13
fix: typescript not using jsx

simllll:
otherwise it breaks parsing some typescript files.


znarf:
Similar to https://github.com/depcheck/depcheck/pull/807

PR #850, opened 2023-11-13
fix: jsx parsing

simllll:
add importAssertions


znarf (approved):
Looks good, this was done in https://github.com/depcheck/depcheck/pull/823 for typescript.

PR #871, opened 2023-12-28
deps: callsite -> callsites

matz3:
The `callsite` package does not include/define a proper license, does not reference a code repository and is not actively maintained.

`callsites` v4 can't be used as it is ESM only.


znarf (approved):
Looks good.

rumpl:
Thanks!

PR #874, opened 2024-01-08
output line in package.json file

### LINKED ISSUE #872: Report package.json line in output
### Feature request description

Actually, the reported output is something like

```
Unused devDependencies
* prettier
```

In the JSON output, we have 

```json
{
  "devDependencies": [
    "prettier"
  ]
}
```

It would be nice to have more detail on how to find "where" is defined this configuration:

For example, the vscode console allow us to click on lines "that looks like a filename with a line number" (ie. `package.json:15`)

Furthermore, some tools like [reviewdog](https://github.com/reviewdog/reviewdog) does parse the output to report checks on github, but actually it's really complicated to target a file. 
Making reviewdog work with depcheck would be really usefull as they are complementary when working with a CI.

A possible nice output would be 

```
Unused devDependencies
* prettier (package.json:15)
```

I think that it's not as easy as said regarding the JSON output, if that's the data structure that you use internally, as we have an array of string here.

The options I see here:

- Do not touch the JSON output, just "improve" the standard output
- Add the package line in the array of string (`"devDependencies": ["prettier (package.json:15)"]`), not pretty, but non-breaking
- change the data format : `"devDependencies": [ { package: "prettier", "dependencyFile": "package.json", "dependencyLine": 15 } ]`

I do not know if the json output is for debbuging purpose or if it is part of public API.

### Any extra info

I can give a hand on this if you are OK with this feature.

jdeniau:
Add the dependency line number in the standard output. See #872 for more about this PR.

Capture d’écran 2024-01-09 à 00 26 36

Fixes #872


rumpl:
Quick question, in terminals that do know how to parse this, is it cmd/ctrl clickable so that the file opens on that line in an IDE?

jdeniau:
Tried in VSCode, it does seems to work great (see video).
Do you have another editor in mind ?

https://github.com/depcheck/depcheck/assets/1398469/8a79b43e-30fb-4e73-8065-6d0e613a2db1

rumpl:
No other in mind no, ok this is great, thanks, now I have to review the code :)

src/cli.js
```diff
@@ -60,11 +61,32 @@ function noIssue(result) {
   );
 }
 
-function prettify(caption, deps, oneline) {
+function prettify(
+  caption,
+  deps,
+  oneline,
+  rootDir = null,
+  printJsonLineInKey = null,
```
jdeniau, line 69: I added two parameters here:
- `rootDir` to open the `package.json`file,
- `printJsonLineInKey` is to find the extra dependency in either "dependencies" or "devDependencies"

src/cli.js
```diff
@@ -60,11 +61,32 @@ function noIssue(result) {
...
+) {
   if (oneline) {
     return deps.length ? [caption, deps.join(' ')] : [];
   }
-  const list = deps.map((dep) => `* ${dep}`);
+
+  const packageJson =
+    rootDir && printJsonLineInKey
+      ? fs.readFileSync(path.join(rootDir, 'package.json'), 'utf8')
+      : null;
```
jdeniau, line 78: Open the package.json if needed (only on default output)

May be optimised to avoid memory consumption with a lot of unused deps (but the purpose of depcheck is to have a few unused deps, so extra optimization may be overkill ?)

src/cli.js
```diff
@@ -60,11 +61,32 @@ function noIssue(result) {
...
+
+  const list = deps.map((dep) => {
+    const lineNumber =
+      packageJson &&
+      printJsonLineInKey &&
+      findLineInFile(packageJson, printJsonLineInKey, dep);
+
+    return `* ${dep}${
+      lineNumber !== null ? ` (package.json:${lineNumber})` : ''
```
jdeniau, line 87: Add the line number if needed

jdeniau (commented):
Great ! I commented on github directly to help you understand the code if needed.

jdeniau:
Hi! Do you need any help from me to get this done?

PR #875, opened 2024-01-09
fix: fix module-root.js edge case where there is a package.json in a subpath

dword-design:
I had an edge case with the `rimraf` package where there was a `package.json` file in the `dist/commonjs` subpath and so depcheck did not detect the rimraf binary. Fixed the issue in this PR and added a test.


rumpl:
@dword-design could you please rebase your PR and fix the red ci? :)

PR #883, opened 2024-02-05
fix: support typescript storybook config

jfairley:
### Description

Storybook can be configured with typescript files, including `main.ts` as an alternative to `main.js`.

This update enables the `storybook` parser for `.storybook/main.ts`.

see https://storybook.js.org/docs/api/main-config


jfairley:
This change along isn't the full solution. The TS config file also allow a default export, which would require further changes to support, but minimally recognizing the `main.ts` file is a relevant step forward.

PR #893, opened 2024-06-21
use jsonc-parser to parse json files

laino:
Parsing JSON files with ``require()`` fails when they contain comments like in this ``tsconfig.json``:

```jsonc
{
  "compilerOptions": {
    /* Visit https://aka.ms/tsconfig to read more about this file */

    /* Projects */
    "incremental": true,                              /* Save .tsbuildinfo files to allow for incremental compilation of projects. */
    "composite": true,                                /* Enable constraints that allow a TypeScript project to be used with project references. */

    /* ... */
  }
}
```

TypeScript will happily read that. The solution is to just use the same relaxed JSON parser TypeScript uses as well.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,6 +22,7 @@
         "is-core-module": "^2.12.0",
         "js-yaml": "^3.14.1",
         "json5": "^2.2.3",
+        "jsonc-parser": "^3.2.0",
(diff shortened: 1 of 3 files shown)


laino:
going to rewrite this since the project is already using json5

PR #895, opened 2024-06-28
Replace `is-core-module` with `@nolyfill/is-core-module`

wojtekmaj:
Node.js from v6.13.0, v8.10.0, v9.3.0 includes `module.builtinModules` which we can use to natively check if some module belongs to Node.js core or not.

This drops not one, but _three_ dependencies, removing 70 KB of bloat: https://npmgraph.js.org/?q=is-core-module

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,6 +11,7 @@
       "dependencies": {
         "@babel/parser": "^7.23.0",
         "@babel/traverse": "^7.23.2",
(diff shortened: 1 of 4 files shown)


ljharb:
Note that this will break for any prefix-only core modules, namely `node:test`, `node:sea`, and `node:sqlite` at this time.

wojtekmaj:
Good call. Thank you @ljharb.

PR #913, opened 2024-12-20
chore: upgrade ESLint

43081j:
Upgrades ESLint and switches to flat config.

This also switches a few plugins, and enables some which previously
weren't applying.

Most code changes are removing unused disable directives, adding a `_`
prefix to unused variables, and switching `forEach` calls to `for...of`.

Depends on #912 since the mocha plugin needs newer mocha


rumpl:
#912 is merged, can you rebase now?

43081j:
done 👍 

one thing worth noting is that `mocha/no-setup-in-describe` triggers fairly often in our codebase

if you'd rather just turn it off globally instead of ignoring each individual case, im happy with that too

we run into it because we do stuff like this:

```ts
describe('xyz', () => {
  // this triggers the lint rule since we're doing logic outside of any hooks
  const fn = someCondition ? 'it' : 'it.only';

  fn('a test', ...);
});
```

43081j:
just trying to figure out some problems only CI seems to be having

we have a mixed repo of CJS (distributed code) and ESM (tests), so something's getting confused

43081j:
ugh

absolute pain but i did get it working in the end

basically, there's a few things to note here...

### disabling node 16

Latest ESLint and/or a plugin we use doesn't support node 16.x, so we'd have to remove it from the CI matrix (it is EOL so this should be ok?)

### node 22.x doesn't like mixed CJS/ESM projects

The repo currently has this structure:

- `src/` - ESM
- `dist/` - compiled CJS
- `test/` - ESM

Because the compiled CJS is what we actually ship, that means this package is _not_ an ES module package (i.e. it should not have `type: "module"` yet).

As that is the case though, it means the files in `test/` should be CJS _or_ they should all have `.mjs` extension

Otherwise, mocha and/or node throws a fit from node 22.x onwards. It does this because it detects we're in a CJS project, but then encounters an ESM file and tries to execute it as such... That then fails since proxyquire is used in some of those files and expects a CJS context.

**Solution** is either of the following:

- migrate everything to ESM (big job, since we rely heavily on using `require(mod)` to load parsers etc)
- migrate tests to CJS (easier, but i tried and it causes many tests to fail)

